PR #68, opened 2022-08-09
tutorial: construction of B-series and average vector field example

ranocha:


docs/src/tutorials/bseries_basics.md
```diff
@@ -144,11 +147,80 @@ And their difference, which is the local error:
...
+
+The average vector field method
+
+```math
+u^{n+1} = u^{n} + \Delta t \int_0^1 f\bigl(\xi u^{n+1} + (1 - \xi) u^{n}\bigr) \mathrm{d} \xi
+```
+
+introduced by McLachlan, Quispel, and Robidoux (1999) is a second-order accurate
+method. Quispel and McLaren (2008) discovered that it is indeed a B-series method.
+It's coefficients are given explicitly as
```
ketch, line 177: This should be "Its" (no apostrophe).

ranocha: Thanks! Fixed in 789b297

docs/src/tutorials/bseries_basics.md
```diff
@@ -144,11 +147,80 @@ And their difference, which is the local error:
...
+
+```math
+\begin{align}
+b(.) &= 1, \\
+b([t_1, ..., t_n]) &= b(t_1)...b(t_n) / (n + 1).
+```
```
ketch, line 183: do we need \end{align} here?

docs/src/tutorials/bseries_basics.md
```diff
@@ -144,11 +147,80 @@ And their difference, which is the local error:
...
+
+by Celledoni. McLachlan, McLaren, Owren, Quispel, and Wright (2009). We can
```
ketch, line 185: period should be comma

ranocha: Thanks! Fixed in c21902a

docs/src/tutorials/bseries_basics.md
```diff
@@ -144,11 +147,80 @@ And their difference, which is the local error:
...
+  "Geometric integration using discrete gradients."
+  Philosophical Transactions of the Royal Society of London.
+  Series A: Mathematical, Physical and Engineering Sciences 357,
+  no. 1754 (1999): 1021-1045.
+- G. Reinout W. Quispel, and David Ian McLaren.
+  "A new class of energy-preserving numerical integration methods."
+  Journal of Physics A: Mathematical and Theoretical 41, no. 4 (2008): 045206.
+- Elena Celledoni, Robert I. McLachlan, David I. McLaren, Brynjulf Owren,
+  G. Reinout W. Quispel, and William M. Wright.
+  "Energy-preserving runge-kutta methods."
```
ketch, line 224: Runge-Kutta

docs/src/tutorials/bseries_basics.md
ranocha: Yes, thanks

docs/src/tutorials/bseries_basics.md
ranocha: Fixed, thanks

ketch:
This looks great.  My only thought is that it might be better to highlight it as a separate standalone example, or else modify the beginning of this example since it says we "investigate error expansions for Runge-Kutta (RK) methods".

ranocha:
Good point. I will fix the build error first and then move it to another tutorial. There, we can briefly present the different methods we have for constructing B-series.

ranocha:
Do we have an interesting application of backward error analysis with the average vector field method? Or maybe a nice illustration of some property we would expect?

ranocha:
I'll merge this for now to make it available. I would of course be happy about feedback - we can improve it later, too.

PR #71, opened 2022-08-15
implement `order_of_accuracy`

ranocha:
I added a function `order_of_accuracy` as convenience function to obtain the order of accuracy of a B-series (used as numerical integrator).

### DIFF
--- a/src/BSeries.jl
+++ b/src/BSeries.jl
@@ -191,6 +191,16 @@ for op in (:*, :\)
     end
 end
 
+# Return a function returning an iterator over all rooted trees used as
+# keys for the B-series when given an order of the trees.
+function _iterator_type(::TruncatedBSeries{<:RootedTree})
+    return RootedTreeIterator
+end
+
+function _iterator_type(::TruncatedBSeries{<:BicoloredRootedTree})
+    return BicoloredRootedTreeIterator
+end
+
 """
     ExactSolution{V}()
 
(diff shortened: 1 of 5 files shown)


ranocha:
This could be extended to something comparing two B-series (optionally up to a given order). However, I couldn't find a good name for something like this - or we would need to use both names `isequal` and `isapprox`. Thus, I decided to provide only  a convenience function `order_of_accuracy` up to now. What do you think about this choice of names, @ketch?

src/BSeries.jl
```diff
@@ -296,6 +299,43 @@ for op in (:+, :-)
...
+    if isempty(kwargs) && !(valtype(series) <: AbstractFloat)
+        compare = isequal
+    else
+        compare = (a, b) -> isapprox(a, b; kwargs...)
+    end
+
+    exact = ExactSolution{valtype(series)}()
+
+    # TODO: TruncatedBSeries
+    # This assumes that the `coef` are always stored with increasing `order` of the
```
ketch, line 324: Why not go with the following approach?  
- Loop over order (starting from 1).
- Generate all trees of that order.
- Check coefficients of each tree.

I guess it would be slower, but generating all trees is still quite fast, right?

ketch:
This seems like a great convenience function.  Thanks!

One thing to note about this is that it makes sense only for B-series that represent a map.  There are other functions that only make sense when applied to maps (e.g. `modified_equation`, and others we may implement in the future).  Do we want to throw an error if the coefficient of the empty tree is not unity?  Or give a warning?  Or modify the function name to contain a hint?  I'm not sure.

ranocha:
> One thing to note about this is that it makes sense only for B-series that represent a map. There are other functions that only make sense when applied to maps (e.g. `modified_equation`, and others we may implement in the future). Do we want to throw an error if the coefficient of the empty tree is not unity? Or give a warning? Or modify the function name to contain a hint? I'm not sure.

Yeah, that's something we should discuss. Right now, we mention it in the docstrings of functions like `compose`. One aspect to think about is how we want to check for this constraint - exactly or approximately with some tolerances? What about symbolic computations?
It could probably be helpful to show a warning or throw an error, but we need to find a robust check before.

src/BSeries.jl
ranocha: Yeah, right. That might just be better. Let me take a look.

ranocha: Done in 9544a9a

ketch (approved):

PR #73, opened 2022-08-18
Rosenbrock methods

ranocha:
Requires https://github.com/SciML/RootedTrees.jl/pull/89


ketch:
Sorry for the delay.  This looks very good.  Since it is mainly an addition to `RootedTrees.jl` I guess it makes sense that it's mainly documented there, but we might want to mention it somewhere in the `BSeries.jl` docs.

It looks like we need to re-run the tests after the other PR is merged.

PR #80, opened 2022-09-07
mention printing settings of RootedTrees.jl in the first tutorial

### LINKED ISSUE #78: Add option to print trees using `butcher_representation()`
Ideally, we would like to print trees in output as actual trees.  This isn't possible, even in a Jupyter notebook, since mathjax doesn't support the necessary packages.  The next best thing is to print the trees using Butcher's notation (\tau's and brackets), which is already implemented in `butcher_representation`.  It would be nice to make this the default, and to have a global option to use level sets instead for those who so desire.

ranocha:
Closes #78

I introduced an option to choose the printing style in https://github.com/SciML/RootedTrees.jl/pull/91. This allows one to set the style on a per-project basis. If you just use the REPL from your main project, you only need to call it once and the setting should be persistent when restarting Julia.

I decided to keep the current style as default since it would be slightly breaking to change it right now. However, we can reconsider that in the future, of course.


ranocha:
@ketch You can see the preview of the docs at https://ranocha.de/BSeries.jl/previews/PR80/tutorials/bseries_basics/

ketch:
This is great.  Thanks!

PR #117, opened 2023-05-17
Add the capability to check whether a B-series is energy-preserving

Sondar74:
This code is based on the Theorem 2 of the paper "Energy-Preserving Integrators and the Structure of B-series" (link: https://link.springer.com/article/10.1007/s10208-010-9073-1).
In this code we want to know if a method is Energy-Preserving or not up to some order. 
The function 'EnergyPreserving(A,b,s)' receives as input a square Matrix which is the inner part of the Butcher Array of a numerical method, a vector b (which length is the same as the amount of rows of A) and an intenger s, which will be the order up to which we are testing this method. 

For example, the classical Runge-Kutta method:
```
A = [0//1  0//1  0//1  0//1
 1//2  0//1  0//1  0//1
 0//1  1//2  0//1  0//1
 0//1  0//1  1//1  0//1]
 b =  [1//6, 1//3, 1//3, 1//6]
```

the function 'EnergyPreserving(A,b,s)' will return 'True' for s = 3,4 but will return 'False' for s = 5, i.e. 'EnergyPreserving(A,b,5)'.

Furthermore, for any method described by the square matrix A and vector b, the function 'OrderMethod' will say up to which order it is energy-preserving. For example, for the same (A,b) as before this function will return "Energy Preserving for order < 4".

Finally, if you have already generated a BSeries `serie` following the next steps:
```
 rka = RungeKuttaMethod(A, b)
#generate bseries 
 series = bseries(rka, s)
```
 the function `is_energy_preserving(serie)` will return `True` for  `s = 4`, and `False` for `s = 5`.


ketch:
Thanks, @Sondar74 !  Could you edit the box above (where it says "no description provided") and add a description of the contents of this PR?  Specifically:

- Describe what functionality is added
- Give example(s) of how to use it

Energy_Preserving.jl
```diff
@@ -0,0 +1,500 @@
+# Load the packages we will use.
+# These must first be installed using: import Pkg; Pkg.add("package_name")
+#import Pkg; Pkg.add("BSeries")
+using BSeries
+using Latexify  # Only needed for some pretty-printing cells below using `latexify`
```
ketch, line 5: Please check which of these packages are actually used, and remove any "using" or "import" statements that are not required.

Energy_Preserving.jl
```diff
@@ -0,0 +1,500 @@
...
+import SymPy; sp=SymPy;
+using LightGraphs
+using Combinatorics
+using RootedTrees
+using LinearAlgebra
+
+#Parametros iniciales
+#stisfies for order <5, and not for 5.
+A = [ 0 0 0
```
ketch, line 14: Let's separate the examples from the main code.  These can be moved to another file (and eventually to the "test" directory, and/or the documentation).

ketch:
Also, it would be great to make a notebook like https://github.com/ranocha/BSeries.jl/blob/main/docs/src/tutorials/bseries_basics.ipynb but explaining and demonstrating the functionality of the code in this PR.  From the Jupyter notebook we can generate a page for the online docs that will look similar to https://ranocha.de/BSeries.jl/dev/tutorials/bseries_basics/.

Energy_Preserving.jl
```diff
@@ -0,0 +1,500 @@
...
+        if IsEnergyPreserving(trees,coefficients) == true
+            flag = true
+        end
+        s = s + 1
+    end
+    println("Energy Preserving for order < ", s-1)
+end
+
+#this function checks if an AVF method is energy Preserving for a given order s
+function EnergyPreservingAVF(s)
```
ranocha, line 349: It would be good to have a general interface of such a function that accepts a B-series. This implementation looks quite similar to `EnergyPreserving(A,b,s)`. If you have a single implementation accepting a B-series, you can re-use it for RK methods and the AVF method (after generating their corresponding B-series).

Energy_Preserving.jl
```diff
@@ -0,0 +1,500 @@
...
+    for i in 1:l
+        factor = 0
+        factor = RootedTrees.symmetry(RootedTree(thetrees[i]))
+        coef[i] = coef[i]*(1//factor)
+    end
+    return coef
+end
+
+#this function tells up to what order a method is Energy Preserving
+function OrderMethod(A,b)
```
ranocha, line 317: This looks strange to me
```julia
julia> # coefficients from your code with comment "stisfies for order <5, and not for 5"
       A = [0 0 0; 1//3 0 0; -5//48 15//16 0]; b = [1//10, 1//2, 2//5]; 

julia> EnergyPreserving(A, b, 1)
All the trees have been checked:
Condition Satisfied
Condition Satisfied

julia> EnergyPreserving(A, b, 2)
All the trees have been checked:
Condition Satisfied
Condition Satisfied

julia> EnergyPreserving(A, b, 3)
All the trees have been checked:
Condition Satisfied
Condition Satisfied

julia> EnergyPreserving(A, b, 4)
All the trees have been checked:
Condition Satisfied
Condition Satisfied

julia> EnergyPreserving(A, b, 5)
All the trees have been checked:
Condition Not Satisfied
Condition Not Satisfied

julia> OrderMethod(A, b)
All the trees have been checked:
Condition Satisfied
Energy Preserving for order < 1
```

ranocha (changes requested):
Thanks for the PR! As @ketch noted, we have a bit of work to do before we can get this into BSeries.jl. Some general comments (besides the aspects already mentioned by @ketch)
- Please follow standard Julia style, e.g., `snake_case` for functions and variables. See https://docs.julialang.org/en/v1/manual/style-guide/
- Please add comments explaining *why* you do something and not only *what* you do
- Please add references to the basic algorithms that you apply
- Please add docstrings

Energy_Preserving.jl
```diff
@@ -0,0 +1,500 @@
...
+    s = 1
+    flag = false
```
ketch, line 319: I think if you rename this flag to something meaningful, it will help.

Energy_Preserving.jl
Sondar74: You are right: the 'if' cycle of the function 'OrderMethod' has to be fixed as I didn´t updated it with the rest of the functions:
if IsEnergyPreserving(trees,coefficients) == false
            flag = true
        end
The only difference is that in the first line I changed the 'true' for false. It should work with this modification.
I apologize for this.

ketch: @Sondar74 You can make changes to the code on your computer, then commit and push them, and this PR will automatically be updated with the new code.

Sondar74:
The Style Guide site recommends not to use the from x -> f(x). Should I rewrite this lines?
`last_j_occurrence = findlast(x -> x == j, a)`

ranocha:
No, that's fine as it is

ranocha (changes requested):

Energy_Preserving.jl
```diff
@@ -0,0 +1,524 @@
...
+# Load the packages we will use.
+using BSeries
+import SymPy; sp=SymPy;
+using Combinatorics
+using RootedTrees
+using LinearAlgebra
+
+
+#this function checks whether a method is energy Preserving for a given order s
+function EnergyPreserving(A,b,s)
```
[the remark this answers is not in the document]
Sondar74: Yes. this function works only for the matrix A and vector B of the Butcher array. If we only need the other one we can get rid of this one.

Energy_Preserving.jl
```diff
@@ -0,0 +1,524 @@
...
+    signal = IsEnergyPreserving(trees,coefficients)  
+    if signal == false
+        println("Condition Not Satisfied")
+    else
+        println("Condition Satisfied")
+    end
+end
+
+#this function checks if the modified_equation of a BSeries is Energy Preserving or not
+function BSeries_Energy_Preserving(bseries)
```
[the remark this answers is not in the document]
Sondar74: Okey, I will fix it into `is_energy_preserving(series,order`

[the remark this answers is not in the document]
Sondar74: As for providing the argument `order`, you want to create another function similar to 'OrderMethod' which returns the order up to which given series are energy-preserving? I mean, the same function but changing the input '(A,b)' (and probably to change the name too).

Energy_Preserving.jl
```diff
@@ -0,0 +1,524 @@
...
+    series_a = modified_equation(bseries)
+    #save all the coefficients in an array
+    coefficients = collect(values(series_a))
+    #save all the RootedTrees in another array: 
+    #we need only the level sequence
+    atrees = collect(keys(series_a))
```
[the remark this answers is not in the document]
Sondar74: In order to work with the index of an array `coefficients[i]`. I don't know is there is another (shorter) way to access into the `values(series_a)` elements.

Energy_Preserving.jl
```diff
@@ -0,0 +1,524 @@
...
+            end
+        end
+    end
+    return modified_t_dict
+end
+
+#this functions checks if the level_sequence is a bush
+function bush_detector(tree)
+    bush = false
+    if tree != [1]
```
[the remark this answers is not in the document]
Sondar74: It works, you are right. 
As for this initial indices, `bseries()` always generates the rooted trees starting with 1, doesn't it?

Energy_Preserving.jl
```diff
@@ -0,0 +1,524 @@
...
+    for j in 1:m
+        ad_dict[j] = diccionario[m-j+1]
+    end
+    return ad_dict
+end
+
+
+#this function returns the adjoint level sequence for a given tree (the input also in level-sequence form)
+#the adoint is calculated with respect to the right-most spine
+function adjoint(a::Vector{Int})
```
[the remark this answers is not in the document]
Sondar74: Okey. Do you think 'energy_preserving_tree' would be a good name? Also `energy_preserving_pair`

Energy_Preserving.jl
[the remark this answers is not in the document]
ranocha: It's fine with me to work on the code of `is_energy_preserving` first

Energy_Preserving.jl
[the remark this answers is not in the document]
ranocha: That's the default behavior but you shouldn't built everything on this assumption

Energy_Preserving.jl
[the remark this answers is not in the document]
ranocha: Finding a good name can often be difficult. If I remember correctly, the basic idea is that you want to look at pairs of trees and the relations of their coefficients to check for energy preservation. Any name indicating this should be fine - but that is not a single tree like `energy_preserving_tree` would indicate, is it?

Energy_Preserving.jl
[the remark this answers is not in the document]
Sondar74: Okey, I'll implement the case for initial index 0

Energy_Preserving.jl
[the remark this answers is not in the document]
Sondar74: I stole the idea from the main paper: it reads in page 678 "[...] where we have replaced each elementary differential by its tree and grouped terms into energy-preserving pairs." 
However, this function returns the `energy_preserving_tree` with respect to the rightmost spine only, but in general there could be more than one tree which satisfy this condition depending on the topology of the tree. Maybe something like `rightmost_energy_preserving_tree`, or if its too long, `rightmost_ep_tree` or `rightmost_energy_preserving`

Energy_Preserving.jl
```diff
@@ -0,0 +1,524 @@
...
+end
+
+#this function multplies the coefficient for its symmetry factor
+function symfact_normalization(coef,thetrees)
+    l = length(coef)
+    for i in 1:l
+        factor = 0
+        #because of the librarys we are using, some packages are repeated
+        #Then, we specify that the symmetry function comes from RootedTrees
+        factor = RootedTrees.symmetry(RootedTree(thetrees[i]))
```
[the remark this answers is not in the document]
Sondar74: You mean to define `import RootedTrees; rt=RootedTrees` so that this line becomes `factor = rt.symmetry(RootedTree(thetrees[i]))`?

Energy_Preserving.jl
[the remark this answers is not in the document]
ranocha: No. The initial index can be any integer

Energy_Preserving.jl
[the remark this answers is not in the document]
ranocha: `rightmost_energy_preserving_tree` is fine with me. Any abbreviation will make it much more unclear what's going on.

Energy_Preserving.jl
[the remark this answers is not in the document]
ranocha: No, the other way round. I guess Combinatorics.jl also exports a `symmetry` - which is the reason why you need to specify `RootedTrees.symmetry` here. What I suggest is to write
```julia
using Combinatorics: whatever, you_need, from_them
```
instead of
```julia
using Combinatorics
```
to import only the names you need from Combinatorics.jl.

Energy_Preserving.jl
ketch: I see that the examples are gone from this file, but they haven't yet been added in a test or documentation.

Energy_Preserving.jl
ketch, line 331: In this line you're reconstructing a `RootedTree` from a level sequence after you (earlier) got the level sequence from the RootedTree.  This (and other similar parts of the code) seem unnecessarily complicated.  I think the code can and should be written using the native objects (RootedTrees and Bseries) as much as possible.  If it doesn't seem possible (or seems overly complicated) for some part of the code, let's discuss it.

ranocha:
I haven't looked at the algorithm in details yet. How does it compare to the approach of Sundklakk (2015)?

Energy_Preserving.jl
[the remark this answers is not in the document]
Sondar74: Okey, I got it.
Then this line becomes `factor = symmetry(RootedTree(thetrees[i]))`

Sondar74: Okey, I made some changes in the main function so that this line becomes only `factor = symmetry(thetrees[i])`

Sondar74:
> I haven't looked at the algorithm in details yet. How does it compare to the approach of Sundklakk (2015)?

I was trying to download the paper from this website, is this the reference you mentioned? And also, can you share the paper with me?

https://www.semanticscholar.org/paper/A-Library-for-Computing-with-Trees-and-B-Series-Sundklakk/4c5c94924f63ed3c61997c5972955d20f83a2145

ranocha:
Done on Slack

Energy_Preserving.jl
```diff
@@ -0,0 +1,521 @@
...
+function add_one_to_left(arrays)
+    l = length(arrays)
+    for i in 1:l
+        arrays[i] = vcat([1], arrays[i])
+    end
+    return arrays
+end
+
+#this function multplies the coefficient for its symmetry factor
+function symfact_normalization(coef,thetrees)
```
ketch, line 321: Let's call this something like `renormalize_bseries`.  Then refactor it to take a BSeries as input and return a dict.

Energy_Preserving.jl
```diff
@@ -0,0 +1,521 @@
...
+    end
+    #normalize the coefficients multiplying by the symmetry factor 
+    coefficients = symfact_normalization(coefficients,trees) 
+    #check if it is energy Preserving 
+    signal = IsEnergyPreserving(trees,coefficients)  
+    return signal
+end
+
+#this function checks if the modified_equation of a BSeries is Energy Preserving or not
+function is_energy_preserving(series)
```
ketch, line 41: Let's combine IsEnergyPreserving with this function into just one.

Energy_Preserving.jl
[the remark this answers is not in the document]
ketch: Let's use multiple dispatch and have two versions of `is_energy_preserving()`: one that accepts a Butcher array and one that accepts a BSeries.

Energy_Preserving.jl
```diff
@@ -0,0 +1,521 @@
...
+        end
+    end
+    #normalize the coefficients multiplying by the symmetry factor 
+    coefficients = symfact_normalization(coefficients,atrees) 
+    #check if it is energy Preserving 
+    signal = IsEnergyPreserving(trees,coefficients)  
+    return signal
+end
+
+function get_leafs(a)
```
ketch, line 66: Let's call this "remove spine".

Project.toml
```diff
@@ -10,6 +10,7 @@ OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
ranocha, line 13: Please add this to your main environment instead of the dependencies of BSeries.jl

src/Energy_Preserving.jl
```diff
@@ -0,0 +1,442 @@
+#This code is based on the Theorem 2 of the paper "Energy-Preserving Integrators and the Structure of B-series" (link: https://link.springer.com/article/10.1007/s10208-010-9073-1).
```
ranocha, line 1: These new files are never `include`d from the main file `src/BSeries.jl`. Thus, their code is not loaded at all.

ranocha (changes requested):

test/runtests.jl
```diff
@@ -1,6 +1,6 @@
 using Test
 using BSeries
-
+include("../src/Energy_Preserving.jl")
```
ranocha, line 3: No. The test code must not `include` code from `src` manually. This has to be handled from within the package

src/BSeries.jl
```diff
@@ -1,5 +1,5 @@
 module BSeries
-
+    include("Energy_Preserving.jl")
```
ranocha, line 2: Please move this below - somewhere where it fits better. For example, you may want to dispatch on things - that need to be defined first

src/BSeries.jl
```diff
@@ -17,6 +17,8 @@ end
 
 using Latexify: Latexify, LaTeXString
 
+using Energy_Preserving
```
ranocha, line 20: It's okay to use a sub-module if you want to keep yhe namespace of BSeries.jl clean but not strictly necessary. Anyway, if you want to introduce a new module, please follow the Julia style guide I linked some time ago.

Project.toml
```diff
@@ -4,12 +4,14 @@ authors = ["Hendrik Ranocha <mail@ranocha.de> and contributors"]
...
 
 [deps]
+Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
ranocha, line 14: Please do not add Revise.jl to the dependencies of BSeries.jl

Energy_Preserving.jl
```diff
@@ -0,0 +1,435 @@
+#This code is based on the Theorem 2 of the paper "Energy-Preserving Integrators and the Structure of B-series" (link: https://link.springer.com/article/10.1007/s10208-010-9073-1).
```
ranocha, line 1: Is this file still needed?

CSRK.jl
```diff
@@ -0,0 +1,83 @@
+#Packages which would be useful eventually
```
ranocha, line 1: What is this file about?

examples.jl
```diff
@@ -0,0 +1,29 @@
+#up to order 4
```
ranocha, line 1: Is this file still needed?

ranocha (changes requested):

src/BSeries.jl
ketch: @ranocha Sorry, @Sondar74 and I have no idea where it would fit better.  We had to hunt for a while to figure out the syntax, and the examples we found put includes at the top.  I also don't understand the comment "you may want to dispatch on things" -- can you clarify?

ketch:
@Sondar74 It may be simpler as @ranocha suggests to simply add your new code to the main bseries.jl file.

The CSRK file is really a different topic and shouldn't be added in this pull request.  We will add it in a future pull request.  You can put it in a different branch (on your machine/fork) and remove it from this branch.  If needed, I can help you with this.

src/BSeries.jl
ranocha: Say you want to have a simplified API such as `is_energy_preserving(rk::RungeKuttaMethod, order)`. To define such a method dispatching on its first argument, `RungeKuttaMethod` must be defined - which comes from RootedTrees.jl. Thus, the definition of this method must come after `using RootedTrees`.

ranocha: A good place may just be at the end of the main file - near the line 
```julia
include("latexify.jl")
```

There, everything you may need for dispatch should be defined.

src/Energy_Preserving.jl
ranocha: If this code is moved to `src/BSeries.jl`, you don't need this file anymore, do you?

src/BSeries.jl
```diff
@@ -1447,4 +1448,437 @@ include("latexify.jl")
...
+This function checks whether a method is energy preserving 
+for a given order s. 
+    Inputs: 
+        A: square matrix
+        b: vector of Butcher array
+        s: integer, the order
+    Output:
+        True/False
+"""
+function is_energy_preserving(A,b,s)
```
ranocha, line 1464: For consistency with the other functions in this package, it would be nice to change this to
```suggestion
function is_energy_preserving(rk::RungeKuttaMethod, order)
```
Then, the implementation should just create a B-series of the appropriate `order` and pass it to `is_energy_preserving(series)`

ranocha, line 1464: You can check the rendering of docstrings locally in the REPL, e.g.,
```julia
julia> """
       ￼        is_energy_preserving(A,b,s)
       ￼
       ￼This function checks whether a method is energy preserving 
       ￼for a given order s. 
       ￼    Inputs: 
       ￼        A: square matrix
       ￼        b: vector of Butcher array
       ￼        s: integer, the order
       ￼    Output:
       ￼        True/False
       ￼"""
       function foo end
foo

help?> foo
search: foo floor pointer_from_objref OverflowError RoundFromZero unsafe_copyto! functionloc StackOverflowError

  ￼ isenergypreserving(A,b,s) ￼ ￼This function checks whether a method is energy preserving ￼for a given order
  s. ￼ Inputs: ￼ A: square matrix ￼ b: vector of Butcher array ￼ s: integer, the order ￼ Output: ￼ True/False ￼
```

src/BSeries.jl
```diff
@@ -1447,4 +1448,437 @@ include("latexify.jl")
...
+
+"""
+        is_energy_preserving(A,b,s)
+
+This function checks whether a method is energy preserving. 
+    Input: 
+        a bseries
+    Output:
+        True/False
+"""
```
ranocha, line 1498:
```suggestion
"""
    is_energy_preserving(series)::Bool

This function checks whether the B-series `series` of a time integration
method is energy preserving for Hamiltonian systems - up to the
[`order`](@ref) of the `series`.

# References

- Please add the main reference here, including a DOI
- You can also add another reference here
"""
```

ranocha (changes requested):

src/BSeries.jl
ketch: It might be nice to allow both interfaces.  But yes, we definitely want the one that takes a RungeKuttaMethod.

Also, let's follow the convention of using "p" for order ("s" is usually the # of stages).

src/BSeries.jl
Sondar74: I did this in the last commit.

src/BSeries.jl
Sondar74: Implemented.

Energy_Preserving.jl
Sondar74: No, its exactly the same as the one located in src. I will delete it

examples.jl
Sondar74: I removed it from this branch.

Project.toml
Sondar74: The line was removed.

src/Energy_Preserving.jl
Sondar74: I removed this file.

Energy_Preserving.jl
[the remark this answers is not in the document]
Sondar74: I generalized it to any integer.

Project.toml
ranocha: Yes, please

Project.toml
ranocha: :+1:

src/BSeries.jl
```diff
@@ -17,6 +17,7 @@ end
 
 using Latexify: Latexify, LaTeXString
 
+
```
ranocha, line 20:
```suggestion

using Combinatorics: permutations

```
I guess that's why you want to use Combinatorics.jl?

Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Hendrik Ranocha <mail@ranocha.de> and contributors"]
 version = "0.1.46"
 
 [deps]
+Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
```
ranocha, line 7: Needs a compat entry

src/BSeries.jl
```diff
@@ -1447,4 +1449,420 @@ include("latexify.jl")
...
+        is_energy_preserving(series)::Bool
+This function checks whether the B-series `series` of a time integration
+method is energy preserving for Hamiltonian systems - up to the
+[`order`](@ref) of the `series`.
+"""
+#References:
+#This code is based on the Theorem 2 of the paper "Energy-Preserving 
+#Integrators and the Structure of B-series".
+# (link: https://link.springer.com/article/10.1007/s10208-010-9073-1).
+function is_energy_preserving(series)
```
ranocha, line 1483: By inserting a comment here, the docstring is not connected to the function definition below.

src/BSeries.jl
Sondar74: Yes, I only need the `permutations` function from it

Sondar74: In fact, I think I also need `rank` from `LinearAlgebra`

src/BSeries.jl
Sondar74: Solved: I quitted the comments and moved the docstring downwards.

Project.toml
Sondar74: I added `Combinatorics = "1.0.2"`

Project.toml
```diff
@@ -33,6 +34,7 @@ SymEngine = "0.8"
 SymPy = "1"
 Symbolics = "4, 5"
 julia = "1.6"
+Combinatorics = "1.0.2"
```
ranocha, line 37:
```suggestion
Combinatorics = "1"
```
Do you really require the patch release 1.0.2? If not, the more general version 1 should be fine

src/BSeries.jl
```diff
@@ -1447,4 +1450,421 @@ include("latexify.jl")
...
+        s: integer, the order
+    Output:
+        True/False
+"""
+function is_energy_preserving(rka::RungeKuttaMethod,p)
+#generate bseries 
+    series = bseries(rka, p)
+    #pass it to the main functoin 'is_energy_preserving'
+    is_energy_preserving(series)
+end
```
ranocha, line 1471:
```suggestion
"""
    is_energy_preserving(rk::RungeKuttaMethod, order)::Bool

This function checks whether a Runge-Kutta method `rk` is 
energy-preserving for a given `order`.
"""
function is_energy_preserving(rk::RungeKuttaMethod, order)
    # generate bseries 
    series = bseries(rk, order)
    # pass it to the main method `is_energy_preserving`
    is_energy_preserving(series)
end
```

src/BSeries.jl
```diff
@@ -1447,4 +1450,421 @@ include("latexify.jl")
...
+        is_energy_preserving(series)::Bool
+This function checks whether the B-series `series` of a time integration
+method is energy preserving for Hamiltonian systems - up to the
+[`order`](@ref) of the `series`.
+References:
+This code is based on the Theorem 2 of the paper "Energy-Preserving 
+Integrators and the Structure of B-series".
+ (link: https://link.springer.com/article/10.1007/s10208-010-9073-1)
+ """
+function is_energy_preserving(series)
```
ranocha, line 1484:
```suggestion
"""
    is_energy_preserving(series)::Bool

This function checks whether the B-series `series` of a time integration
method is energy-preserving for Hamiltonian systems - up to the
[`order`](@ref) of the `series`.

# References

This code is based on the Theorem 2 of
- Elena Celledoni, Robert I. McLachlan, Brynjulf Owren, and G. R. W. Quispel. 
  "Energy-preserving integrators and the structure of B-series."
  Foundations of Computational Mathematics 10 (2010): 673-693.
  [DOI: 10.1007/s10208-010-9073-1](https://link.springer.com/article/10.1007/s10208-010-9073-1)
 """
function is_energy_preserving(series)
```

src/BSeries.jl
```diff
@@ -1447,4 +1450,421 @@ include("latexify.jl")
...
+    series_a = modified_equation(series)
+    #save all the coefficients in an array
+    coefficients = collect(values(series_a))
+    #save all the RootedTrees in another array: 
+    atrees = collect(keys(series_a))
```
ranocha, line 1489: Please write comments that explain *why* you do something - *what* you do should become clear from the code itself.

ranocha (changes requested):

Project.toml
Sondar74: Done

src/BSeries.jl
Sondar74: Implemented.

src/BSeries.jl
```diff
@@ -1453,4 +1456,420 @@ include("latexify.jl")
...
+
+This function checks whether a Runge-Kutta method `rk` is 
+energy-preserving for a given `order`.
+"""
+function is_energy_preserving(rk::RungeKuttaMethod,order)
+#generate bseries 
+    series = bseries(rk, order)
+    #pass it to the main functoin 'is_energy_preserving'
+    is_energy_preserving(series)
+end
```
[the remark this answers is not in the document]
ranocha: You can click on "Commit suggestion" to apply a suggestion as above directly. This avoids copy-paste errors that I assume have happened...

src/BSeries.jl
```diff
@@ -1453,4 +1456,420 @@ include("latexify.jl")
...
+    series_a = modified_equation(series)
+    #save all the coefficients in an array: it is easier
+    #to call an array full of fractions when working with the 
+    #functions 'renormalize_bseries' and 'energy_preserving_trees_test' 
+    coefficients = collect(values(series_a))
+    #save all the RootedTrees in another array: 
+    atrees = collect(keys(series_a))
+    #we need only the level sequence
+# Create an empty vector to store the converted trees into arrays
+    trees = Vector{Vector{Int}}(undef, length(series_a))
```
ranocha, line 1497: Why do you work with a vector of level sequences instead of a vector of rooted trees?

src/BSeries.jl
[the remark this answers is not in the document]
Sondar74: Thank you, I'll do it next time

src/BSeries.jl
```diff
@@ -1453,4 +1456,420 @@ include("latexify.jl")
...
+# Convert the trees and store them in the 'trees' vector
+    for i in 1:length(series_a)
+        levelsequence = atrees[i].level_sequence
+        if isempty(levelsequence)
+            trees[i] = Int[]
+        else
+            trees[i] = levelsequence
+        end
```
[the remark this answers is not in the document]
Sondar74: You are right. I'll fix it

src/BSeries.jl
Sondar74: I chose to work with the level sequences because we have to modify the numbers inside a tree (for example, in the function `modify_t_sub`), and I found it easier by using arrays instead of Rooted Trees

Project.toml
```diff
@@ -33,6 +34,7 @@ SymEngine = "0.8, 0.9.1"
 SymPy = "1"
 Symbolics = "4, 5"
 julia = "1.6"
+Combinatorics = "1"
```
ranocha, line 37: The [compat] section should be ordered alphabetically - this line needs to be moved a bit

src/BSeries.jl
```diff
@@ -1453,4 +1456,419 @@ include("latexify.jl")
...
+function eliminate_repeated_subarrays(M)
+    unique_M = []
+    for arr in M
+        if arr in unique_M
+        else
+            push!(unique_M, arr)
+        end
+    end
+    return unique_M
+end
```
ranocha, line 1615:
```julia
help?> unique!
search: unique! unique allunique
[...]
  unique!(A::AbstractVector)

  Remove duplicate items as determined by isequal, then return the modified A. unique! will return the elements
  of A in the order that they occur. If you do not care about the order of the returned data, then calling
  (sort!(A); unique!(A)) will be much more efficient as long as the elements of A can be sorted.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> unique!([1, 1, 1])
  1-element Vector{Int64}:
   1
  
  julia> A = [7, 3, 2, 3, 7, 5];
  
  julia> unique!(A)
  4-element Vector{Int64}:
   7
   3
   2
   5
  
  julia> B = [7, 6, 42, 6, 7, 42];
  
  julia> sort!(B);  # unique! is able to process sorted data much more efficiently.
  
  julia> unique!(B)
  3-element Vector{Int64}:
    6
    7
   42
```

src/BSeries.jl
```diff
@@ -1453,4 +1456,419 @@ include("latexify.jl")
...
+function indexator(trees,onetree)
+    theindex = 0
+    l = length(trees)
+    for i in 1:l
+        if trees[i] == onetree
+            theindex = i
+        end
+    end
+    return theindex
+end
```
ranocha, line 1679:
```julia
help?> findfirst
search: findfirst
[...]
  findfirst(predicate::Function, A)

  Return the index or key of the first element of A for which predicate returns true. Return nothing if there
  is no such element.

  Indices or keys are of the same type as those returned by keys(A) and pairs(A).

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> A = [1, 4, 2, 2]
  4-element Vector{Int64}:
   1
   4
   2
   2
  
  julia> findfirst(iseven, A)
  2
  
  julia> findfirst(x -> x>10, A) # returns nothing, but not printed in the REPL
  
  julia> findfirst(isequal(4), A)
  2
  
  julia> A = [1 4; 2 2]
  2×2 Matrix{Int64}:
   1  4
   2  2
  
  julia> findfirst(iseven, A)
  CartesianIndex(2, 1)
```
?

src/BSeries.jl
```diff
@@ -1453,4 +1456,419 @@ include("latexify.jl")
...
+            #save the index of the level_sequence: this will be used for creating the matrix EP
+            highindex = indexator(trees,t)
+            #check if the level_sequence corresponds to a bush
+            bushflag = bush_detector(t)
+            #if it does, then the coefficient must be equal to zero.
+            if bushflag == true
+                if coefficients[highindex] != 0
+                    error_searcher = true
+                end
+            else
```
ranocha, line 1819: This could just be
```
            # Check whether the level sequence corresponds to a bushy tree.
            # If so, we can exit early since the coefficients of bushy trees must be
            # zero in the modified equation of energy-preserving methods.
            if bush_detector(t)
                return false
            end
```

src/BSeries.jl
```diff
@@ -1453,4 +1456,419 @@ include("latexify.jl")
...
+
+"""
+This function generates all the equivalent trees 
+for a given level_sequence
+"""
+function equivalent_trees(array)
+    tree = rootedtree(array)
+    l = length(array)
+    #we get the permutations of the array
+    superarray = get_permutations(array)
```
ranocha, line 1692: This is the main bottleneck:
```julia
julia> using ProfileView, BSeries

julia> series = bseries(8) do t, series
           if order(t) in (0, 1)
               return 1 // 1
           else
               v = 1 // 1
               n = 0
               for subtree in SubtreeIterator(t)
                   v *= series[subtree]
                   n += 1
               end
               return v / (n + 1)
           end
       end

julia> @profview is_energy_preserving(series) # result of calling this the second time, after compilation
```

image

When you refine the implementation, you should consider concentrating on this. 
- Instead of saving all permutations first and throwing away most of them later, could you just iterate over all permutations lazily and just keep the ones that are required?
- Is there another algorithm with reduced complexity to generate all equivalent representation of a rooted tree?

src/BSeries.jl
Sondar74: Implemented

src/BSeries.jl
Sondar74: Implemented

src/BSeries.jl
Sondar74: I modified the comments and the structure of this function.

src/BSeries.jl
```diff
@@ -1453,4 +1456,419 @@ include("latexify.jl")
...
+    not_energy_preserving = false
+    while not_energy_preserving == false 
+#generate bseries 
+        if is_energy_preserving(rka,p+1) == false
+            not_energy_preserving = true
+        end
+        p = p + 1
+    end
+    return p-1
+end
```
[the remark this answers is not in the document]
ketch: This function (at least the version that accepts a BSeries as input) does need some way to stop in case you give it an exactly energy-preserving method like AVF.  Perhaps there could be a default maximum order, and if it reaches that it prints a message saying the method is EP to *at least* that order.  And the user could override the default if they want to check to higher order.

Sondar74:
> @Sondar74 I realized later yesterday that instead of "spine" and "ribs", it probably would make more sense to use "trunk" and "branches" (you had already suggested branches, which got me thinking about this).

Celledoni also calls it "spine" in the page 6 (678), during the first line of the proof of the Teorem 2

src/BSeries.jl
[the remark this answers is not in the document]
Sondar74: I added this parameter

ketch:
> Celledoni also calls it "spine" in the page 6 (678), during the first line of the proof of the Teorem 2

Oh, that must be where I got the word from; I had forgotten.  Nevertheless, I think it would be nice to use "trunk" and "branches" as they are more consistent with other terminology used in this area.

src/BSeries.jl
```diff
@@ -1453,4 +1456,379 @@ include("latexify.jl")
...
+# References
+This code is based on the Theorem 2 of
+- Elena Celledoni, Robert I. McLachlan, Brynjulf Owren, and G. R. W. Quispel. 
+  "Energy-preserving integrators and the structure of B-series."
+  Foundations of Computational Mathematics 10 (2010): 673-693.
+  [DOI: 10.1007/s10208-010-9073-1](https://link.springer.com/article/10.1007/s10208-010-9073-1)
+ """
+function is_energy_preserving(map_series)
+    flow_series = modified_equation(map_series)
+    #check if the first element is map or flow.
```
ketch, line 1489: I think you meant to add an "assert" here to check that the input is indeed a map.

src/BSeries.jl
```diff
@@ -1453,4 +1456,379 @@ include("latexify.jl")
...
+        return false
+    end
+    #loop over same-order trees
+    order = 5
+    while order <= max_length && energy_preserving_flag == true
+        same_order_trees = []
+        same_order_coeffs = []
+        #we create and fill the arrays of trees and coefficients
+        #corresponding to an 'order'
+        for j in 1:l
```
ketch, line 1522: Instead of this, why not just use `RootedTreeIterator(order)`?

Then also you can get the coefficients via `flow_series[t]` where `t` is a tree.  So then I think there is no need to make the separate arrays of trees and coefficients.

src/BSeries.jl
```diff
@@ -1453,4 +1456,379 @@ include("latexify.jl")
...
+end
+
+#        remove_trunk(a)
+
+#This function returns the forest of level_sequences 
+#obtained after removing the rightmost trunk of a given
+#level_sequence.
+function remove_trunk(a)
+    m = num_branchs(a)
+    branchs_array = Array{Array}(undef, m)
```
ketch, line 1564: It's clear that this is an array, but that doesn't help with readability.  Let's just call it `branches`.

ketch: Also, "branches" needs to have an "e" here and everywhere that it appears.

src/BSeries.jl
```diff
@@ -1453,4 +1456,379 @@ include("latexify.jl")
...
+            #n+m-2j+1 for every number in the level_sequence
+        else
+            branchs[j] = [n+m-2j+1 for n in branchs_array[j]]
+        end
+    end
+    return reverse(branchs)
+end
+
+
+#This functions checks if the level_sequence is a bush.
```
ketch, line 1611: Should be "This function" (no "s" at the end).

src/BSeries.jl
```diff
@@ -1453,4 +1456,379 @@ include("latexify.jl")
...
+    return superarray
+end
+
+
+function num_branchs(a)
+    k = a[end]
+    return k - 1
+end
+
+function add_one_to_left(arrays)
```
ketch, line 1714: Please add a description for this method.

src/BSeries.jl
```diff
@@ -1453,4 +1456,379 @@ include("latexify.jl")
...
+        arrays[i] = vcat([1], arrays[i])
+    end
+    return arrays
+end
+
+"""
+        renormalize_bseries!(series)
+
+This function receives a 'TruncatedBSeries' and returns
+a new 'TruncatedBseries' which coefficients are multiplied by
```
ketch, line 1726: I think this should say "whose coefficients are divided by the symmetry of the corresponding tree."

Sondar74: Implemented.

src/BSeries.jl
Sondar74: Implemented.

src/BSeries.jl
Sondar74: Implemented.

src/BSeries.jl
Sondar74: Implemented.

src/BSeries.jl
Sondar74: Implemented

src/BSeries.jl
Sondar74: Sorry, I don't see how we can use `RootedTreeIterator(order)` here

src/BSeries.jl
```diff
@@ -1485,8 +1485,9 @@ This code is based on the Theorem 2 of
   [DOI: 10.1007/s10208-010-9073-1](https://link.springer.com/article/10.1007/s10208-010-9073-1)
  """
 function is_energy_preserving(map_series)
+    #check if the first element is map or flow
+    @assert typeof(map_series) == TruncatedBSeries{RootedTree{Int64, Vector{Int64}}, Rational{Int64}}
```
ketch, line 1489: This just checks that the input is a BSeries.  What we want to check is whether the BSeries represents a map or a flow, based on whether the first coefficient is 1 or zero.  We discussed this last week in my office.

src/BSeries.jl
```diff
@@ -1554,61 +1555,61 @@ function low_order_energy_preserving(trees, coefficients)
...
-#        remove_trunk(a)
+#        branches(a)
 
 #This function returns the forest of level_sequences 
 #obtained after removing the rightmost trunk of a given
 #level_sequence.
-function remove_trunk(a)
-    m = num_branchs(a)
-    branchs_array = Array{Array}(undef, m)
+function branches(a)
```
ketch: It's a bit confusing to me that you have two functions named `branches` and `get_branches`.  Especially since `branches` does seem to get the branches while `get branches` actually moves the branches around.  Please add a comment explaining what `get_branches` does and then consider renaming or merging these functions so that the names are clear.

src/BSeries.jl
ketch: Oh, I see that my comment wasn't clear.  I didn't mean to change the name of the function `remove_trunk`, but instead to change the name of `branches_array` to `branches`.

src/BSeries.jl
ketch: You need all the trees of a specific order, right?  `RootedTreeIterator(order)` gives you exactly that.

ketch:
```
same_order_trees = collect(RootedTreesIterator(order))
same_order_coeffs = [flow_series[tree] for tree in same_order_trees]
```

src/BSeries.jl
ketch: I guess if you change `branches` back to `remove_trunk` then it resolves this issue.

src/BSeries.jl
Sondar74: Okey. I changed `branches_array` for `branches.` On the other hand, I linked the functions `get_branches` and `modify_branches` into only one function called `get_branches`.

src/BSeries.jl
Sondar74: I merged the two functions into one `get_branches` function. I hope this is easier to read

src/BSeries.jl
```diff
@@ -1485,9 +1485,9 @@ This code is based on the Theorem 2 of
   [DOI: 10.1007/s10208-010-9073-1](https://link.springer.com/article/10.1007/s10208-010-9073-1)
  """
 function is_energy_preserving(map_series)
-    #check if the first element is map or flow
-    @assert typeof(map_series) == TruncatedBSeries{RootedTree{Int64, Vector{Int64}}, Rational{Int64}}
     flow_series = modified_equation(map_series)
+    #check if the first element is map or flow
+    @assert flow_series[rootedtree([1])] == 1
```
ranocha, line 1490:
```suggestion
    if !isone(flow_series[rootedtree([1])])
        throw(ArgumentError("This method is designed for B-series representing a flow."))
    end
```

src/BSeries.jl
```diff
@@ -1556,6 +1556,46 @@ function low_order_energy_preserving(trees, coefficients)
...
+by the symmetry of the corresponding tree.
+"""
+function renormalize_bseries!(series)
+    trees_array = collect(keys(series))
+    l = length(trees_array)
+    for i in 1:l
+        factor = symmetry(trees_array[i])
+        series[trees_array[i]] = series[trees_array[i]]*(1//factor)
+    end
+end
```
ranocha, line 1573:
```suggestion
"""
        renormalize_bseries!(series)

This function modifies a B-series by dividing each coefficient
by the `symmetry` of the corresponding tree.
"""
function renormalize_bseries!(series)
    for t in keys(series)
        series[t] /= symmetry(t)
    end
    return series
end
```

ranocha (changes requested):

src/BSeries.jl
Sondar74: Implemented.

src/BSeries.jl
ketch: @ranocha What would you suggest?  Perhaps after all this function should just return an array of coefficients, to make it clear that they do not represent a B-Series in the same way that the `BSeries` object does?

ketch: Or we could do away with this function entirely and just divide by the symmetry when needed?  I suppose that is what is done in other parts of the code, like when we print a B-Series?

Sondar74: Okey, so we don't want to modify the original BSeries but only to save the renormalized version in another variable and work with this last one, right?

ranocha: I'm not sure, to be honest. This is an ugly problem. As far as I see, we have the following options.

1. Just use this method as it is and tell everyone that they need to be careful.
2. Return an array of normalized coefficients - but we lose the direct association with trees in this case.
3. Add another field to the existing implementation(s) of B-series like `normalized::Bool` defaulting to `true`. Check this field whenever we perform operations on the B-series and throw an error if required. Change this field to `false` when `renormalize!` is called. The function `renormalize!` could switch between both modes.
4. Tell people that they need to renormalize the coefficients themselves. That's essentially what we are doing right now, e.g., in

https://github.com/ranocha/BSeries.jl/blob/04b2f999612dea844e05febea33dfc2dbac6165d/src/BSeries.jl#L361L366

ranocha (changes requested):
Thanks! The current version is already much more efficient :+1: 

Next, there are still some type instabilities and further possibilities for improvements. You can get an impression of the overall performance profile via
```julia
julia> using ProfileView, BSeries

julia> series = bseries(10) do t, series
           if order(t) in (0, 1)
               return 1 // 1
           else
               v = 1 // 1
               n = 0
               for subtree in SubtreeIterator(t)
                   v *= series[subtree]
                   n += 1
               end
               return v / (n + 1)
           end
       end

julia> @profview is_energy_preserving(series) # ignore first profile including compilation

julia> @profview is_energy_preserving(series) # use second profile after compilation
```
This yields
image
for me. Some takeaways are
1. A substantial amount of time (roughly 50%) is spent in `modified_equation` - which is fine.
2. You can see that there are type instabilities - the red bars. They are for example cause by lines such as `same_order_trees = []` and same_order_coeffs = []`, where you create `Vector{Any}`, fill them, and pass them to another function. Something like this should be avoided by specifying the types concretely (based on the coefficients you get).
3. You use dense linear algebra. I think it may be worth a try to use sparse linear algebra instead. Then, you should also avoid collecting all the column vectors by just collecting the relevant values and indices while assembling the sparse matrix.
4. You call `rank` twice. Depending on whether this is a bottleneck at the end, I wonder whether there is a clever way to avoid this using the SVD or something similar directly.

src/BSeries.jl
ketch: @ranocha My code above doesn't work, and I don't quite understand why.  Can you tell us what is the right way to get an array from the iterator?

Screen Shot 2023-06-13 at 11 43 47 AM

src/BSeries.jl
ketch: How about we go with option 1 for now, but create an issue to continue discussing this?

src/BSeries.jl
```diff
@@ -1453,4 +1456,399 @@ include("latexify.jl")
...
+    for i in 1:length(flow_series)
+        trees[i] = thetrees[i].level_sequence
+    end
+    energy_preserving_flag = true
+    max_length = length(trees[end])
+    l = length(trees)
+    #first, check if the energy_preserving condition is satisfied
+    #low order (provided that the computational cost at order < 5)
+    # is low. This is because the 'rank' function's behavior
+    # in the test function doesn't works by separate for low orders.
```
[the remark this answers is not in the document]
ketch: Regarding performance, I think perhaps what we have is good enough for now.  We can raise an issue with suggestions for future performance improvements.  Let me know what you think @ranocha .

src/BSeries.jl
ranocha: The tree `t` is a view to an internal buffer for efficiency. You can use something like
```julia
julia> map(copy, RootedTreeIterator(4))
4-element Vector{RootedTree{Int64, Vector{Int64}}}:
 RootedTree{Int64}: [1, 2, 3, 4]
 RootedTree{Int64}: [1, 2, 3, 3]
 RootedTree{Int64}: [1, 2, 3, 2]
 RootedTree{Int64}: [1, 2, 2, 2]
```

See also https://github.com/SciML/RootedTrees.jl#iteration-over-rootedtrees and https://sciml.github.io/RootedTrees.jl/stable/api_reference/#RootedTrees.RootedTreeIterator

src/BSeries.jl
ranocha: Yes, sounds good to me :+1:

src/BSeries.jl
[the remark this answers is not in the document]
ranocha: Sure, we can do that.

src/BSeries.jl
ranocha: https://github.com/ranocha/BSeries.jl/issues/129

src/BSeries.jl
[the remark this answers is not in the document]
ranocha: https://github.com/ranocha/BSeries.jl/issues/130

ranocha:
I will make some fixes to this PR and perform a final review before merging the first working version

src/BSeries.jl
ranocha, line 1512: I don't understand this comment. Could you please explain it to me?

src/BSeries.jl
ketch: Thanks!

src/BSeries.jl
ketch: @Sondar74 edited this earlier today but I guess he hasn't pushed it.  @Sondar74 please push your changes so @ranocha can do a final review.

ranocha: Thanks!

ranocha (approved):
Thanks a lot!

PR #119, opened 2023-05-23
order Project.toml to let Aqua.jl tests pass

ranocha:


ranocha:
Tests indicate that the Project.toml should be changed but the new version is exactly what we have here. Thus, I will merge this to see whether it's confused by the different ordering on `main` and check again

ranocha:
I really have no idea what's wrong here - all tests pass locally both with Julia v1.6.7 and v1.9.0.

PR #134, opened 2023-06-13
provide convenience function for the AVF method

ranocha:
I think it would be nice to have a convenience function returning the B-series of the AVF method. However, I'm not sure about the best interface. We could have

1. `bseries(AverageVectorFieldMethod(), up_to_order)`
2. `average_vector_field_method(up_to_order)`

with options to allow different types of the coefficients, e.g.,

1. `bseries(AverageVectorFieldMethod(Float64), up_to_order)`
2. `average_vector_field_method(Float64, up_to_order)`

Option 1 would be more close to most methods of `bseries` we have so far - but the return value of `AverageVectorFieldMethod()` would have no further use (at the moment). Option 2 is more close to something like `zeros`.

What would you prefer, @ketch?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "BSeries"
 uuid = "ebb8d67c-85b4-416c-b05f-5f409e808f32"
 authors = ["Hendrik Ranocha <mail@ranocha.de> and contributors"]
-version = "0.1.49"
+version = "0.1.50"
 
 [deps]
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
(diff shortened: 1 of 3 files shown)


ketch:
I have asked @Sondar74 to develop some capabilities for CSRK methods (of which AVF2 is an example).  These methods can be fully represented by a combination of:

- A polynomial basis
- A (real-valued) matrix

So it might make sense to implement your 1st approach, where `AverageVectorFieldMethod()` returns the matrix (and possibly a field indicating which polynomial basis it relates to).  The next step is to write a function that computes the elementary weights from this representation, and @Sondar74 is working on that now.

ranocha:
Great! I will implement the first approach here so that we have the convenience method `bseries(AverageVectorFieldMethod(), up_to_order)` soon.

ranocha:
I've updated this PR. It would be nice if you could have a look, @ketch

ketch:
This looks fine.  Once @Sondar74 has written the more general version, we may consider replacing this one that is a particular case (or perhaps we will keep it, since the 2nd-order AVF method is particularly interesting).

PR #137, opened 2023-06-14
add some additional tests of `is_energy_preserving` (including currently broken ones)

ranocha:


ketch:
Sympy does have a `rank` function that seems to work with symbolic coefficients.  As far as I can tell, it assumes the values of the coefficients don't lie in any measure-zero subset that would reduce the rank.

ranocha:
That should be fine and would be a great addition for another PR :+1:

PR #141, opened 2023-06-21
Adding CSRK code to BSeries.jl

Sondar74:
This code is based on the paper "A characterization of energy-preserving methods and the construction of parallel integrators for Hamiltonian systems" (https://doi.org/10.48550/arXiv.1505.02537).

This code requires to import SymPy in order to work with symbolic variables for the polynomials generation and the integration process. The BSeries.jl was modified by adding a function which generates a polynomial for a square matrix, and another function `elementary_differentials_csrk` which returns the elementary differential for a given square matrix and a RootedTree. Only this last function is being exported. An exmaple of its functionality can be tested as follows:

```
#define a matrix M
M = [-6//5 72//5 -36 24;
    72//5 -144//5 -48 72;
    -36 -48 720 -720;
    24 72 -720 720]
rt = rootedtree([1,2,3])
elementary_differentials_csrk(M,rt)
```


ranocha:
You can use backticks (i.e., `) to mark code regions. Instead of

´´´
#define a matrix M
M = [-6//5 72//5 -36 24;
72//5 -144//5 -48 72;
-36 -48 720 -720;
24 72 -720 720]
rt = rootedtree([1,2,3])
elementary_differentials_csrk(M,rt)
´´´

you would get

```julia
#define a matrix M
M = [-6//5 72//5 -36 24;
72//5 -144//5 -48 72;
-36 -48 720 -720;
24 72 -720 720]
rt = rootedtree([1,2,3])
elementary_differentials_csrk(M,rt)
```

by writing this as

````
```julia
#define a matrix M
M = [-6//5 72//5 -36 24;
72//5 -144//5 -48 72;
-36 -48 720 -720;
24 72 -720 720]
rt = rootedtree([1,2,3])
elementary_differentials_csrk(M,rt)
```
````

**Comment by ranocha** [comment]
Please resolve the conflicts shown in 
image

**Comment by ranocha** [comment]
Before we go into th details of the implementation, let's talk about the interface. Most existing interfaces of BSeries.jl to create a B-series are of the form
```julia
series = bseries(method, order)
```
Thus, it would be nice if you could create a new `struct` to parameterize a CSRK method. Then, you can add another method of `bseries` to compute the coefficients. 

Do you have another suggestion, @ketch ?

Sondar74:
> Please resolve the conflicts shown in image

Sorry, how can I eliminate the extra files from the PR?

Sondar74:
> Before we go into th details of the implementation, let's talk about the interface. Most existing interfaces of BSeries.jl to create a B-series are of the form
> 
> ```julia
> series = bseries(method, order)
> ```
> 
> Thus, it would be nice if you could create a new `struct` to parameterize a CSRK method. Then, you can add another method of `bseries` to compute the coefficients.
> 
> Do you have another suggestion, @ketch ?

Okey. I'll start working on this `struct`. Should the interface be something like `series = csrkseries(matrix, order)`?

ranocha:
> Okey. I'll start working on this `struct`. Should the interface be something like `series = csrkseries(matrix, order)`?

I would prefer something along the lines of
```julia
csrk = ContinuousStageRungeKuttaMethod(matrix)
series = bseries(csrk, order)
```

ketch:
Sorry, I'm in a place with extremely limited internet access this week.  Hopefully this comment gets transmitted.

@Sondar74 we had discussed the question of whether the computation of elementary differentials belongs in BSeries.jl or RootedTrees.jl.  @ranocha What is your opinion?

Then there is the question of where the code that defines the structure of a CSRK method should go.  If we follow the pattern of `RungeKuttaMethod`, that should also go in `RootedTrees`, right?

ranocha:
We have defined most of these schemes in RootedTrees.jl, but that was also for historical reasons. I'm fine keeping more involved time integration schemes here in BSeries.jl. We can always move the to RootedTrees.jl if we want.
The reason is that you can really use the functionality of RootedTrees.jl to derive order conditions of RK methods without needing B-series theory. But something like CSRK methods is really interesting mainly a theoretical point of view, so it's fine to have them here where we have more advanced theoretical tools to analyze them.

ketch:
Okay, that seems reasonable to me.

ketch:
Okay, that seems reasonable to me.

Sondar74:
Closing Obsolete PR

PR #144, opened 2023-06-26
Implementing CSRK code in BSeries.jl

Sondar74:
This code is based on the paper "A characterization of energy-preserving methods and the construction of parallel integrators for Hamiltonian systems" (https://doi.org/10.48550/arXiv.1505.02537).

Added a struct in BSeries.jl called ContinuousStageRungeKutta and a function CSRK().

# Example:
The energy-preserving 4x4 matrix given by Miyatake & Butcher (2015) is
```
M = [-6//5   72//5  -36//1  24//1;
        72//5  -144//5 -48//1  72//1;
        -36//1 -48//1   720//1 -720//1;
        24//1   72//1  -720//1  720//1]
```

Then, we calculate the bseries with the following code:

```
csrk = ContinuousStageRungeKuttaMethod(M)
series = bseries(csrk, 4)
TruncatedBSeries{RootedTree{Int64, Vector{Int64}}, Rational{Int64}} with 9 entries:
  RootedTree{Int64}: Int64[]      => 1//1
  RootedTree{Int64}: [1]          => 1//1
  RootedTree{Int64}: [1, 2]       => 1//2
  RootedTree{Int64}: [1, 2, 3]    => 1//6
  RootedTree{Int64}: [1, 2, 2]    => 1//3
  RootedTree{Int64}: [1, 2, 3, 4] => 1//24
  RootedTree{Int64}: [1, 2, 3, 3] => 1//12
  RootedTree{Int64}: [1, 2, 3, 2] => 1//8
  RootedTree{Int64}: [1, 2, 2, 2] => 1//4

```

### DIFF
--- a/src/BSeries.jl
+++ b/src/BSeries.jl
@@ -632,7 +632,7 @@ in order to later call the 'bseries' function.
   SIAM Journal on Numerical Analysis 54, no. 3 (2016): 
   [DOI: 10.1137/15M1020861](https://doi.org/10.1137/15M1020861) 
 """
-struct ContinuousStageRungeKuttaMethod{T, MatT <: AbstractMatrix{T}} <: RootedTrees.AbstractTimeIntegrationMethod
+struct ContinuousStageRungeKuttaMethod{MatT <: AbstractMatrix} <: RootedTrees.AbstractTimeIntegrationMethod
     A::MatT
 end
 
@@ -664,7 +664,7 @@ M = [-6//5   72//5  -36//1  24//1;
 Then, we calculate the B-series with the following code:
 
 ```
-csrk = CSRK(M)
+csrk = ContinuousStageRungeKuttaMethod(M)
 series = bseries(csrk, 4)
 TruncatedBSeries{RootedTree{Int64, Vector{Int64}}, Rational{Int64}} with 9 entries:
   RootedTree{Int64}: Int64[]      => 1//1
@@ -686,7 +686,7 @@ TruncatedBSeries{RootedTree{Int64, Vector{Int64}}, Rational{Int64}} with 9 entri
   [DOI: 10.1137/15M1020861](https://doi.org/10.1137/15M1020861) 
 """
 function bseries(csrk::ContinuousStageRungeKuttaMethod, order)
-    csrk = csrk.matrix
+    csrk = csrk.A
     V_tmp = eltype(csrk)
     if V_tmp <: Integer
         # If people use integer coefficients, they will likely want to have results


ranocha:
The failing tests come from a deprecation warning fixed in https://github.com/JuliaPy/PyCall.jl/pull/1042.

ranocha (changes requested):
Thanks a lot for the PR! Now, everything seems to work. Please close the obsolete PRs.

- Please add references and explain why you are doing what you are doing.
- Please add tests.
- Please check type stability (also in the tests).
- Could you please describe an example of the new functionality in your opening post? 
- How is the performance behavior compared to the other implementations generating B-series we have so far?

src/BSeries.jl
```diff
@@ -40,6 +40,8 @@ export renormalize!
 
 export is_energy_preserving, energy_preserving_order
 
+export elementary_differentials_csrk, CSRK
```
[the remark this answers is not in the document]
Sondar74: Okey, I'll withdraw this.

src/BSeries.jl
```diff
@@ -612,6 +628,23 @@ function bseries(ros::RosenbrockMethod, order)
     return series
 end
 
+"""
+bseries CSRK
+"""
+function bseries(csrk::ContinuousStageRungeKuttaMethod, order)
+    csrk = csrk.matrix
+    V = Rational{Int64}
+    series = TruncatedBSeries{RootedTree{Int, Vector{Int}}, V}()
```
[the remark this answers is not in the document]
Sondar74: Do you refer to this function?
``` 
function bseries(::AverageVectorFieldMethod{T}, o) where {T}
    bseries(o) do t, series
        if order(t) in (0, 1)
            return one(T)
        else
            v = one(T)
            n = 0
            for subtree in SubtreeIterator(t)
                v *= series[subtree]
                n += 1
            end
            return v / (n + 1)
        end
    end
end
```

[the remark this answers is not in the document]
ranocha: Yes.

src/BSeries.jl
```diff
@@ -74,6 +76,20 @@ end
...
+    matrix::MatT
+end
+
+function CSRK(matrix::AbstractMatrix)
+    T = promote_type(eltype(matrix))
+    _M = T.(matrix)
+    return ContinuousStageRungeKuttaMethod(_M)
+end
+
+
```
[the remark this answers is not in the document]
Sondar74: If I delete these lines, is it still possible to generate the object `ContinuousStageRungeKuttaMethod`?

src/BSeries.jl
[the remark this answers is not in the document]
Sondar74: Implemented.

src/BSeries.jl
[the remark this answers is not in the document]
ranocha: Yes. That's calling the default constructor you are already calling here as `ContinuousStageRungeKuttaMethod(_M)`

src/BSeries.jl
```diff
@@ -40,6 +40,8 @@ export renormalize!
 
 export is_energy_preserving, energy_preserving_order
 
+export CSRK
```
ranocha, line 43:
```suggestion
export ContinuousStageRungeKuttaMethod
```
That's how the struct is called below. Currently, we don't use abbreviations, so I would like to keep it like this. Please open an issue for discussing whether we may want to export common abbreviations such as `RK`, `AVF`, or `CSRK` in the future.

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
+"""
+struct ContinuousStageRungeKuttaMethod{MatT <: AbstractMatrix}
+    matrix::MatT
+end
+
+function CSRK(matrix::AbstractMatrix)
+    T = promote_type(eltype(matrix))
+    _M = T.(matrix)
+    return ContinuousStageRungeKuttaMethod(_M)
+end
```
ranocha, line 634:
```suggestion
```
See above

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
+
+"""
+        bseries(csrk::ContinuousStageRungeKuttaMethod, order)
+
```
ranocha, line 638:
```suggestion
"""
    bseries(csrk::ContinuousStageRungeKuttaMethod, order)

```
Indentation with 4 space to mark it as code. With 8 spaces, it will be indented code as in

```
  bseries(ros::RosenbrockMethod, order)

  Compute the B-series of the Rosenbrock method ros up to a prescribed integer order.

  │ Normalization by elementary differentials
  │
  │  The coefficients of the B-series returned by this method need to be multiplied by a power of the
  │  time step divided by the symmetry of the rooted tree and multiplied by the corresponding
  │  elementary differential of the input vector field f. See also evaluate.

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────

      bseries(csrk::ContinuousStageRungeKuttaMethod, order)

  Return a truncated B-series up to the specified order with coefficients determined by the square matrix
  located in csrk. The coefficients for every RootedTree are obtained via the 'elementarydifferentialscsrk'
  function, which is calculated according to Miyatake & Butcher (2015). [See @ref csrk]
```

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
+Return a truncated B-series up to the specified `order` with coefficients
+determined by the square matrix located in `csrk`. The coefficients for every 
+RootedTree are obtained via the 'elementary_differentials_csrk' function,
+which is calculated according to Miyatake & Butcher (2015). [See @ref csrk]
```
ranocha, line 642:
```suggestion
Compute the B-series of the [`ContinuousStageRungeKuttaMethod`](@ref) `csrk`
up to the prescribed integer `order` as described by Miyatake & Butcher (2015).

!!! note "Normalization by elementary differentials"
    The coefficients of the B-series returned by this method need to be
    multiplied by a power of the time step divided by the `symmetry` of the
    rooted tree and multiplied by the corresponding elementary differential
    of the input vector field ``f``.
    See also [`evaluate`](@ref).
```
- This suggestion is adapted to the format we use for other methods.
- If you mention `elementary_differentials_csrk` in a docstring, it becomes part of the public API. That's something I suggested above not to do (by not `export`ing it).
- The syntax `[See @ref csrk]` is wrong.

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
+  RootedTree{Int64}: [1]          => 1//1
+  RootedTree{Int64}: [1, 2]       => 1//2
+  RootedTree{Int64}: [1, 2, 3]    => 6004799503160661//36028797018963968
+  RootedTree{Int64}: [1, 2, 2]    => 6004799503160661//18014398509481984
+  RootedTree{Int64}: [1, 2, 3, 4] => 6004799503160661//144115188075855872
+  RootedTree{Int64}: [1, 2, 3, 3] => 6004799503160661//72057594037927936
+  RootedTree{Int64}: [1, 2, 3, 2] => 1//8
+  RootedTree{Int64}: [1, 2, 2, 2] => 1//4
+
+```
```
ranocha, line 669: If you rewrite this in form of a jldoctest, it will be tested in CI automatically.

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
+# References
+Butcher, John & Miyatake, Yuto. (2015). A Characterization of Energy-Preserving
+Methods and the Construction of Parallel Integrators for Hamiltonian Systems. 
+SIAM Journal on Numerical Analysis. 54. 10.1137/15M1020861.
```
ranocha, line 673:
```suggestion
# References

- Yuto Miyatake and John C. Butcher.
  "A characterization of energy-preserving methods and the construction of
  parallel integrators for Hamiltonian systems."
  SIAM Journal on Numerical Analysis 54, no. 3 (2016): 
  [DOI: 10.1137/15M1020861](https://doi.org/10.1137/15M1020861) 
```
- This uses the same style as other reference comments and includes a link for the DOI
- The order of authors of the current version does not match the order at SIAM

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
 end
 
+"""
+    ContinuousStageRungeKuttaMethod
+
+A struct that describes a CSRK method. This kind of 'struct' should be constructed 
+via [`CSRK`] 
+        'csrk = CSRK(M)' 
+in order to later call the 'bseries' function. 
+"""
```
ranocha, line 625: It would be nice to include at least a reference to the paper of Miyatake and Butcher here, too.

src/BSeries.jl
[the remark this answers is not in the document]
Sondar74: Okey, so I must export `ContinuousStageRungeKuttaMethod`, correct?

src/BSeries.jl
```diff
@@ -612,6 +615,74 @@ function bseries(ros::RosenbrockMethod, order)
...
+function bseries(csrk::ContinuousStageRungeKuttaMethod, order)
+    csrk = csrk.matrix
+    bseries(o) do t, series
+        if order(t) in (0, 1)
+            return one(csrk)
+        else
+            return elementary_differentials_csrk(csrk, t)
+        end
+    end
+end
```
ranocha, line 684: Please add tests. This does not work correctly right now.

ranocha (changes requested):

src/BSeries.jl
Sondar74: Okey, I'm working on this.

Sondar74: When working with the function `one()`, I have the problem that `eltype(csrk)` returns `Any`. How can I fix it?

ranocha: What is the `csrk` that you use here?

ranocha (changes requested):
- Please add tests with floating point coefficients.
- Please add tests with symbolic coefficients using SymPy.jl, SymEngine.jl, and Symbolics.jl.

src/BSeries.jl
Sondar74: It is the `csrk` obtained from `csrk = ContinuousStageRungeKuttaMethod(M)`, for a matrix `M`.

src/BSeries.jl
```diff
@@ -612,6 +616,93 @@ function bseries(ros::RosenbrockMethod, order)
...
+csrk = CSRK(M)
+series = bseries(csrk, 4)
+TruncatedBSeries{RootedTree{Int64, Vector{Int64}}, Rational{Int64}} with 9 entries:
+  RootedTree{Int64}: Int64[]      => 1//1
+  RootedTree{Int64}: [1]          => 1//1
+  RootedTree{Int64}: [1, 2]       => 1//2
+  RootedTree{Int64}: [1, 2, 3]    => 6004799503160661//36028797018963968
+  RootedTree{Int64}: [1, 2, 2]    => 6004799503160661//18014398509481984
+  RootedTree{Int64}: [1, 2, 3, 4] => 6004799503160661//144115188075855872
+  RootedTree{Int64}: [1, 2, 3, 3] => 6004799503160661//72057594037927936
```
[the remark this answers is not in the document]
Sondar74: No, sorry: seems like there is something happening with the definition of `V` in 
```
series = TruncatedBSeries{RootedTree{Int, Vector{Int}}, V}()
series[rootedtree(Int[])] = one(V)
``` 
so that, although the output of `elementary_differentials_csrk(csrk, t)` is correct, for some values it will throw an approximation to the original result instead of saving the value of `elementary_differentials_csrk(csrk, t)` Do you know how can I fix it?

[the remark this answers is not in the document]
Sondar74: Seems like the numerical function `N()` solves this issue, but I don't know if it has limitations for future applications of this code (since this is the first time I'm working with this function).

src/BSeries.jl
ranocha: Use `eltype(csrk.matrix)` instead or define `eltype(csrk::ContinuousStageRungeKuttaMethod)` appropriately.

src/BSeries.jl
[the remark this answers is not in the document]
ranocha: Let's see. First, please add more tests etc. Then, possible problem can show up already

Sondar74:
> * Please add tests with floating point coefficients.
> * Please add tests with symbolic coefficients using SymPy.jl, SymEngine.jl, and Symbolics.jl.

Seems like a symbolic matrix with inputs from `SymEngine` is not compatible with the `PolynomialsA` function.

ranocha:
> > * Please add tests with floating point coefficients.
> > * Please add tests with symbolic coefficients using SymPy.jl, SymEngine.jl, and Symbolics.jl.
> 
> Seems like a symbolic matrix with inputs from `SymEngine` is not compatible with the `PolynomialsA` function.

That doesn't surprise me since SymPy.jl is probably not compatible with SymEngine.jl etc.

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            # Define variables
+            a = -0.37069987
+            b = 0.74139974
+            c = 2.51720052
+
+            # Create symbolic matrix
+            Minv = [a 1//2 b 1//6;
+            1//2 1//4 1//6 1//8;
+            b 1//6 c 1//10;
+            1//6 1//8 1//10 1//12]
```
ranocha, line 2094: Please check your comments everywhere. This is not a symbolic matrix since you set the coefficients above to fixed floating point values. Where did you get these special choices from? Please add some references

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            order = 4
+            series = bseries(csrk, order)
+            expected_coefficients = [1.0 ,1.0, 0.5000000000000002, 0.16666666666666666, 0.3333333333333336, 0.04166666666666661, 0.0833333333333332, 0.12500000000000003, 0.2500000000000003]
+            @test collect(values(series)) == expected_coefficients
+        end
+        @testset "Symbolic coefficients using SymPy.jl" begin
+            # Define variables
+            import SymPy: symbols
+            x = symbols("x")
+            y = symbols("y")
```
ranocha, line 2109:
```suggestion
            x = SymPy.symbols("x")
            y = SymPy.symbols("y")
```
Please check the other tests how to write them. In particular, please do not `import` stuff inside of testsets.

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            # Create symbolic matrix
+            M = [x y;
+                y x]
+            csrk = ContinuousStageRungeKuttaMethod(M)
+            # Generate the bseries up to order 4
+            order = 3
```
ranocha, line 2115: Again, please check your comments.

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            series = bseries(csrk, order)
+            expected_coefficients = [1,1, 25*x^2/32 + 5*x*y/4 + y^2/2,  1105*x^3/2304 + 671*x^2*y/576 + 545*x*y^2/576 + 37*y^3/144, 125*x^3/192 + 25*x^2*y/16 + 5*x*y^2/4 + y^3/3]
```
ranocha, line 2117: Did you compute this by hand? How did you arrive at these coefficients?

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            @test collect(values(series)) == expected_coefficients
+        end
+        @testset "Symbolic coefficients using SymEngine.jl" begin
+            # Define variables
+            import SymEngine: symbols
```
ranocha, line 2122:
```suggestion
```
See above

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            # Generate the bseries up to order 4
+            order = 3
+            series = bseries(csrk, order)
+            expected_coefficients = [1,1, 25*x^2/32 + 5*x*y/4 + y^2/2,  1105*x^3/2304 + 671*x^2*y/576 + 545*x*y^2/576 + 37*y^3/144, 125*x^3/192 + 25*x^2*y/16 + 5*x*y^2/4 + y^3/3]
+            @test collect(values(series)) == expected_coefficients
+        end
+        @testset "Symbolic coefficients using Symbolics.jl" begin
+            # Define variables
+            import Symbolics: @variables
+            @variables x y
```
ranocha, line 2137:
```suggestion
            Symbolics.@variables x y
```
See above

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            # Create symbolic matrix
+            M = [x y;
+                y x]
+            csrk = ContinuousStageRungeKuttaMethod(M)
+            # Generate the bseries up to order 4
+            order = 3
```
ranocha, line 2143: Comments...

ranocha (changes requested):

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            72//5  -144//5 -48//1  72//1;
+            -36//1 -48//1   720//1 -720//1;
+            24//1   72//1  -720//1  720//1]
+            csrk = ContinuousStageRungeKuttaMethod(M)
+            # Generate the bseries up to order 4
+            order = 4
+            series = bseries(csrk, order)
+            expected_coefficients = [1//1 ,1//1, 1//2, 1//6, 1//3, 1//24, 1//12, 1//8, 1//4]
+            @test collect(values(series)) == expected_coefficients
+        end
```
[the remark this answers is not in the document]
Sondar74: Yes, if I am not wrong it whould be energy preserving for any order. 
I checked it with `is_energy_preserving` and it returns `true`. 
Do you want me to add this as a test?

test/runtests.jl
Sondar74: Okey, I'll fix the comments. As for the values, I took them from the project I am working on with @ketch,  I calculated them

test/runtests.jl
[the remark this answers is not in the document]
ranocha: Yes, please

test/runtests.jl
```diff
@@ -2058,4 +2058,92 @@ using Aqua: Aqua
...
+            1//6 1//8 1//10 1//12]
+
+            # Calculate inverse of the  matrix
+            M = inv(Minv)
+            csrk = ContinuousStageRungeKuttaMethod(M)
+            # Generate the bseries up to order 4
+            order = 4
+            series = bseries(csrk, order)
+            expected_coefficients = [1.0 ,1.0, 0.5000000000000002, 0.16666666666666666, 0.3333333333333336, 0.04166666666666661, 0.0833333333333332, 0.12500000000000003, 0.2500000000000003]
+            @test collect(values(series)) == expected_coefficients
```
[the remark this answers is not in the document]
Sondar74: I think the `is_energy_preserving` function will not necesary work accurately for `Floating` coefficients: for example, if we wanted to express `1/3` in floating we may need to cut it in some point at some decimal place, so there would be loss of information and `is_energy_preserving` would return `false`

test/runtests.jl
Sondar74: No, I just copied from the output of the working code

test/runtests.jl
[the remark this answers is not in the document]
ranocha: Did you test it?

test/runtests.jl
ranocha: Hmm... So it is just a regression test - assuming that the current implementation is correct. Could you please also add real tests using examples with known behavior? IIRC, Miyatake & Butcher (2015) have some examples with free coefficients in their paper. You could express these free coefficients as symbolic variables and check whether the resulting B-series of the CSRK method 

- has the correct `order_of_accuracy`
- `is_energy_preserving`

PR #152, opened 2023-07-31
Fix 'is_energy_preserving' for non-zero 4th-order Rooted Trees

Sondar74:
The tests we made before for the function `is_energy_preserving` work for any Rooted Tree of any order. However, the coefficients of the RootedTrees of order 4 in the modified equations are equal to zero, and the code fails for non-zero coefficients. 

In order to solve this issue, I added a function `fourth order exception`, so that for Rooted Trees of order 4 this function will check if the Energy-Preserving condition postulated by (Celledoni et all, 2010) is satisfied for the trees `[1,2,3,2]`, `[1,2,2,2]` and `[1,2,3,3]`.
Also added a test "Non-zero 4th-order coefficients" for this case.

Link to Celledoni's paper:
https://www.researchgate.net/publication/220104368_Energy-Preserving_Integrators_and_the_Structure_of_B-series


ketch:
I don't understand -- what is the bug being fixed here?  And why do we need to write special code for specific trees?  It would be much nicer if we had a single formula that applied for all trees (since the theorem also is stated that way).

PR #156, opened 2023-08-16
Tolerance for is_energy_preserving Floating Inputs

Sondar74:
Since the code verifies exact equalities, I modified the function `is_energy_preserving_dense` adding a tolerance to the `rank` so that RK methods given by a numerical Butcher array can be detected to be energy-preserving according to a tolerance.


src/BSeries.jl
```diff
@@ -1916,12 +1916,13 @@ function _is_energy_preserving_dense(trees, coefficients)
...
         isempty(t) && continue
-
         # Check whether the level sequence corresponds to a bushy tree.
         # If so, we can exit early since the coefficients of bushy trees
         # must be zero in the modified equation of energy-preserving methods.
         if length(t) > 1 && is_bushy(t)
-            if !iszero(coefficients[t_index])
+            # Set the tolerance 
+            if !isapprox(coefficients[t_index],0; atol=1e-10)
+                println("flag")
```
ranocha, line 1925: Please clean up such debug statements

src/BSeries.jl
```diff
@@ -1952,16 +1953,18 @@ function _is_energy_preserving_dense(trees, coefficients)
 
     # We remove repeated columns
     unique!(energy_preserving_basis)
+    # Set the numerical tolerance
+    tolerance = 1e-8
 
     # The components of `energy_preserving_basis` are the columns of the matrix `M`.
     M = reduce(hcat, energy_preserving_basis)
-    rank_M = rank(M)
+    rank_M = rank(M, atol = tolerance)
```
ranocha, line 1961: If we use a tolerance here, we should discuss whether it makes sense to allow users passing the tolerance as keyword arguments.

src/BSeries.jl
```diff
@@ -1952,16 +1953,18 @@ function _is_energy_preserving_dense(trees, coefficients)
...
 
     # We also create an extended matrix for which we append the vector of
     # coefficients
     Mv = [M coefficients]
-    rank_Mv = rank(Mv)
-
+    rank_Mv = rank(Mv, atol = tolerance)
```
ranocha, line 1966: Why do you choose an absolute tolerance here instead of the default relative tolerance?

ranocha (changes requested):

src/BSeries.jl
ketch: Yes, the tolerance should definitely be passed in by the user.  We should probably default to zero tolerance.

src/BSeries.jl
Sondar74: Okey.

src/BSeries.jl
```diff
@@ -1916,12 +1916,13 @@ function _is_energy_preserving_dense(trees, coefficients)
...
         # We do not need to consider the empty tree
         isempty(t) && continue
-
         # Check whether the level sequence corresponds to a bushy tree.
         # If so, we can exit early since the coefficients of bushy trees
         # must be zero in the modified equation of energy-preserving methods.
         if length(t) > 1 && is_bushy(t)
-            if !iszero(coefficients[t_index])
+            # Set the tolerance 
+            if !isapprox(coefficients[t_index],0; atol=1e-10)
```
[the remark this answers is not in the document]
Sondar74: I was testing some methods with the code, but I will set the same tolerance for both `rank` and the part of the Bushy Trees.

src/BSeries.jl
Sondar74: For some methods obtained from Butcher arrays with floating inputs some information may be lost so that  the coefficients may not be exactly equal to the coefficient of their `energy_preserving_pair`. Then, for these methods it is better to set a tolerance to test if they are `energy-preserving` or not.

src/BSeries.jl
ranocha: The default setting for `rank` includes a non-zero relative tolerance:

```julia
julia> using LinearAlgebra

help?> rank
search: rank RankDeficientException lowrankupdate lowrankupdate! lowrankdowndate lowrankdowndate! readlink

  rank(A::AbstractMatrix; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ)
  rank(A::AbstractMatrix, rtol::Real)

  Compute the rank of a matrix by counting how many singular values of A have magnitude greater than max(atol,
  rtol*σ₁) where σ₁ is A's largest singular value. atol and rtol are the absolute and relative tolerances,
  respectively. The default relative tolerance is n*ϵ, where n is the size of the smallest dimension of A, and
  ϵ is the eps of the element type of A.

  │ Julia 1.1
  │
  │  The atol and rtol keyword arguments requires at least Julia 1.1. In Julia 1.0 rtol is available as
  │  a positional argument, but this will be deprecated in Julia 2.0.
```

I think it does not really make sense to use vanishing tolerances for the computation of the `rank` of matrices in floating point precision. I don't have a good suggestion but the kind of default used there looks reasonable to me.

src/BSeries.jl
ranocha: Yes, I totally agree with this. What I was asking is whether it would make more sense to use a relative tolerance by default instead of an absolute one.

Sondar74: Oh, I see. I didn't think about it. 
I can implement the relative tolerance instead of the absolute one if you prefer.

src/BSeries.jl
ketch: Yes, you're right -- @Sondar74 and I had discussed it and come to the same conclusion.

src/BSeries.jl
```diff
@@ -1774,6 +1774,34 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
...
+
+This function checks up to which order a Runge-Kutta method `rk` is
+energy-preserving for Hamiltonian problems, with a parameter of Tolerance. 
+Default tolerance is atol = 0.
+It requires a `max_order` so that it does not run forever if the order up to
+which the method is energy-preserving is too big or infinite.
+
+See also [`is_energy_preserving`](@ref)
+"""
+function energy_preserving_order(rk::RungeKuttaMethod, max_order, tolerance)
```
ranocha, line 1788: We already have `function energy_preserving_order(rk::RungeKuttaMethod, max_order)` above. Please modify the existing functionality and do not introduce another method.

src/BSeries.jl
```diff
@@ -1800,7 +1842,7 @@ This code is based on the Theorem 2 of
   Foundations of Computational Mathematics 10 (2010): 673-693.
   [DOI: 10.1007/s10208-010-9073-1](https://link.springer.com/article/10.1007/s10208-010-9073-1)
 """
-function is_energy_preserving(series_integrator)
+function is_energy_preserving(series_integrator, tolerance)
```
ranocha, line 1845: 1. The change needs to be reflected in the docstring
2. I would again prefer optional keyword arguments with sensible default values

ranocha (changes requested):
Please also write some new tests checking the new functionality accordingly

Sondar74:
I have a question about the `runtest` file: in the test `Example in Section 4.2 of Miyatake and Butcher (2016)`, what is the purpose of the command `@inferred`?

src/BSeries.jl
```diff
@@ -1754,10 +1754,13 @@ This function checks up to which order a Runge-Kutta method `rk` is
...
 It requires a `max_order` so that it does not run forever if the order up to
 which the method is energy-preserving is too big or infinite.
+Keyword Arguments:
+- `atol::Float64=1e-14`: The absolute tolerance for energy preservation.
+- `rtol::Float64=1e-14`: The relative tolerance for energy preservation.
 
 See also [`is_energy_preserving`](@ref)
 """
-function energy_preserving_order(rk::RungeKuttaMethod, max_order)
+function energy_preserving_order(rk::RungeKuttaMethod, max_order; atol::Float64=1e-14, rtol::Float64=1e-14)
```
ranocha, line 1763: We should not restrict the types so severely. Doing it like this means that we will not really be able to use other floating point precisions, for example.

ranocha:
Please note that `rank` for sparse matrices uses a library that requires different keyword arguments

```julia
julia> using SparseArrays, LinearAlgebra

help?> rank
search: rank RankDeficientException lowrankupdate lowrankupdate! lowrankdowndate lowrankdowndate! readlink

  rank(A::AbstractMatrix; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ)
  rank(A::AbstractMatrix, rtol::Real)

  Compute the rank of a matrix by counting how many singular values of A have magnitude greater than max(atol,
  rtol*σ₁) where σ₁ is A's largest singular value. atol and rtol are the absolute and relative tolerances,
  respectively. The default relative tolerance is n*ϵ, where n is the size of the smallest dimension of A, and
  ϵ is the eps of the element type of A.

  │ Julia 1.1
  │
  │  The atol and rtol keyword arguments requires at least Julia 1.1. In Julia 1.0 rtol is available as
  │  a positional argument, but this will be deprecated in Julia 2.0.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> rank(Matrix(I, 3, 3))
  3
  
  julia> rank(diagm(0 => [1, 0, 2]))
  2
  
  julia> rank(diagm(0 => [1, 0.001, 2]), rtol=0.1)
  2
  
  julia> rank(diagm(0 => [1, 0.001, 2]), rtol=0.00001)
  3
  
  julia> rank(diagm(0 => [1, 0.001, 2]), atol=1.5)
  1

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────

  rank(::QRSparse{Tv,Ti}) -> Ti

  Return the rank of the QR factorization

  ────────────────────────────────────────────────────────────────────────────────────────────────────────────

  rank(S::SparseMatrixCSC{Tv,Ti}; [tol::Real]) -> Ti

  Calculate rank of S by calculating its QR factorization. Values smaller than tol are considered as zero. See
  SPQR's manual.

```

This leads to the CI failure https://github.com/ranocha/BSeries.jl/actions/runs/5917130538/job/16045150066?pr=156#step:7:450

ranocha:
There are also other CI errors, e.g.,

```
Floating64: Error During Test at /home/runner/work/BSeries.jl/BSeries.jl/test/runtests.jl:2174
  Test threw exception
  Expression: energy_preserving_order2(rk, 10, atol = 0.0001) == 5
  UndefVarError: `energy_preserving_order2` not defined
```

at https://github.com/ranocha/BSeries.jl/actions/runs/5917130538/job/16045150066?pr=156#step:7:540

Sondar74:
> ```julia
> using SparseArrays, LinearAlgebra
> ```

Do you think we should set the default absolute tolerance to 0?

ranocha:
> Do you think we should set the default absolute tolerance to 0?

I'm not sure about it. The default of 
```Julia
rank(A::AbstractMatrix; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ)
```
looks reasonable to me.

Sondar74:
> > Do you think we should set the default absolute tolerance to 0?
> 
> I'm not sure about it. The default of
> 
> ```julia
> rank(A::AbstractMatrix; atol::Real=0, rtol::Real=atol>0 ? 0 : n*ϵ)
> ```
> 
> looks reasonable to me.

Okey. I plan to use an `if` cycle in the function `_is_energy_preserving_sparse` to check if the user specified a tolerance or if we must use the default one. Is it fine?

ranocha:
You can for example set the default value of the keyword argument for the tolerance to `nothing`. Then, you can use an `if` statement as you said, checkin `isnothing`.

src/BSeries.jl
```diff
@@ -1754,10 +1754,13 @@ This function checks up to which order a Runge-Kutta method `rk` is
...
 It requires a `max_order` so that it does not run forever if the order up to
 which the method is energy-preserving is too big or infinite.
+Keyword Arguments:
+- `atol::Real=0`: The absolute tolerance for energy preservation.
+- `rtol::Real=0`: The relative tolerance for energy preservation.
 
 See also [`is_energy_preserving`](@ref)
 """
-function energy_preserving_order(rk::RungeKuttaMethod, max_order)
+function energy_preserving_order(rk::RungeKuttaMethod, max_order; atol::Real=0, rtol::Real=0)
```
ranocha, line 1763: This should be adapted to the convention we will use for `is_energy_preserving`

src/BSeries.jl
```diff
@@ -1775,22 +1778,32 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
 end
 
 """
-    is_energy_preserving(rk::RungeKuttaMethod, order)::Bool
-
```
ranocha, line 1779: Please do not delete this part of the docstring - but adapt it to the new calling signature

src/BSeries.jl
```diff
@@ -1775,22 +1778,32 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
...
 This function checks whether the Runge-Kutta method `rk` is
 energy-preserving for Hamiltonian systems up to a given `order`.
+Keyword Arguments:
+- `atol::Real=0`: The absolute tolerance for energy preservation.
+- `rtol::Real=0`: The relative tolerance for energy preservation.
+
+# Arguments
+- `rk::RungeKuttaMethod`: The Runge-Kutta method to be evaluated.
+- `order::Int`: The order up to which energy preservation is checked.
```
ranocha, line 1789: This should be adapted to the method accepting a B-series

src/BSeries.jl
```diff
@@ -1886,9 +1899,9 @@ end
 # of coefficients.
 # Checks whether `trees` and `ocefficients` satisfify the energy_preserving
 # condition.
-function _is_energy_preserving(trees, coefficients)
+function _is_energy_preserving(trees, coefficients, atol, rtol)
     # TODO: `Float32` would also be nice to have. However, the default tolerance
-    #       of the rank computation is based on `Float64`. Thus, it will usually
+    #       of the rank computation is based on `Real`. Thus, it will usually
```
ranocha, line 1904: Why did you change this?

src/BSeries.jl
```diff
@@ -1897,32 +1910,45 @@ function _is_energy_preserving(trees, coefficients)
...
         isempty(t) && continue
-
         # Check whether the level sequence corresponds to a bushy tree.
         # If so, we can exit early since the coefficients of bushy trees
         # must be zero in the modified equation of energy-preserving methods.
         if length(t) > 1 && is_bushy(t)
-            if !iszero(coefficients[t_index])
-                return false
+            if atol == 0 && rtol == 0
+                if !isapprox(coefficients[t_index],0)
```
ranocha, line 1937: `isapprox(SOMETHING, 0)` will always fail - unless the first argument is exactly zero. For example,
```julia
julia> isapprox(1.0e-300, 0)
false
```

Thus, using `isapprox` here is confusing, I think

src/BSeries.jl
```diff
@@ -1937,7 +1963,7 @@ function _is_energy_preserving_dense(trees, coefficients)
                 # Check if the rightmost energy-preserving tree is in the
                 # set of trees
                 idx = findfirst(isequal(energy_preserving_pair), trees)
-                if idx !== nothing # i.e., energy_preserving_pair in trees
+                if idx !== 0 # i.e., energy_preserving_pair in trees
```
ranocha, line 1966: Why did you change this?

src/BSeries.jl
```diff
@@ -1952,24 +1978,24 @@ function _is_energy_preserving_dense(trees, coefficients)
 
     # We remove repeated columns
     unique!(energy_preserving_basis)
-
```
ranocha, line 1955: Please do not introduce whitespace changes in unrelated parts of the code

src/BSeries.jl
```diff
@@ -1952,24 +1978,24 @@ function _is_energy_preserving_dense(trees, coefficients)
...
     M = reduce(hcat, energy_preserving_basis)
     rank_M = rank(M)
 
     # We also create an extended matrix for which we append the vector of
     # coefficients
     Mv = [M coefficients]
+    
     rank_Mv = rank(Mv)
-
+
```
ranocha, line 1990: Please do not introduce whitespace changes in unrelated parts of the code

src/BSeries.jl
```diff
@@ -1897,32 +1910,45 @@ function _is_energy_preserving(trees, coefficients)
...
+                end
+            elseif atol != 0 && rtol == 0
+                if !isapprox(coefficients[t_index],0; atol=atol)
+                    return false
+                end
+            else
+                if !isapprox(coefficients[t_index],0; atol=atol, rtol=rtol)
+                    return false
+                end
             end
```
ranocha, line 1952: This looks very complicated to me - and contains quite a lot of branches that make it a bit difficult to check. Moreover, a relative tolerance alone does nothing when comparing something with zero (see above).

src/BSeries.jl
```diff
@@ -1991,8 +2017,22 @@ function _is_energy_preserving_sparse(trees, coefficients)
...
+                    return false
+                end
+            elseif atol != 0 && rtol == 0
+                if !isapprox(coefficients[t_index],0; atol=atol)
+                    return false
+                end
+            else
+                if !isapprox(coefficients[t_index],0; atol=atol, rtol=rtol)
+                    return false
+                end
```
ranocha, line 2035: See above

src/BSeries.jl
```diff
@@ -2008,7 +2048,7 @@ function _is_energy_preserving_sparse(trees, coefficients)
                 # Check if the rightmost energy-preserving tree is in the
                 # set of trees
                 idx = findfirst(isequal(energy_preserving_pair), trees)
-                if idx !== nothing # i.e., energy_preserving_pair in trees
+                if idx !== 0 # i.e., energy_preserving_pair in trees
```
ranocha, line 2051: ?

src/BSeries.jl
```diff
@@ -2024,8 +2064,15 @@ function _is_energy_preserving_sparse(trees, coefficients)
 
     # The components of `energy_preserving_basis` are the columns of the matrix `M`.
     M = sparse(rows, cols, vals, length(coefficients), maximum(cols))
-    rank_M = rank(M)
+    # Compute the singular value decomposition (SVD) of the sparse matrix
+    U, Σ, V = svd(Matrix(M))
```
ranocha, line 2068: No. We explicitly used sparse matrices for efficiency. You convert them here to dense matrices. That defeats the purpose of this way of computing things and will likely make it quite a lot slower for bigger systems.

ranocha (changes requested):
Do we need to be able to adapt the tolerances for the `rank` computations at all - or is it sufficient to introduce a tolerance for the comparisons with zero? In the latter case, we could just introduce a single keyword argument (maybe `tol`?) with default value determined by the `eltype` of the -series. Maybe something like this
```
V = eltype(series_integrator)
if V <: AbstractFloat
  tol = 100 * eps(T)
else
  tol = zero(T)
end
```
What do you think about this?

src/BSeries.jl
Sondar74: So that it accepts both `Int` and `Float64`.

src/BSeries.jl
Sondar74: I think it works well for `int` inputs, but for `Float` it is necessary to set a Tolerance using the keywords.

src/BSeries.jl
Sondar74: Sorry, this was a mistake.

src/BSeries.jl
Sondar74: Yes, sorry.

src/BSeries.jl
Sondar74: Ah, sorry. I'll fix it back.

src/BSeries.jl
ranocha: What do you mean by

> I think it works well for `int` inputs

Could you please show me an example?

Sondar74: For example, the `modified_equation` of a B-Series with integer inputs will have some coefficients `0//1`. In this case, `isapprox()` will accept it without specifying a tolerance.
`julia> isapprox(0, 0//1)
true`

ranocha: Yes - since `0//1` is exactly zero. That can already be checked by `iszero` or `== 0`. `isapprox` is usually meant to check equality with some tolerance. However, that does not work with the default tolerances when checking whether something is zero.

Sondar74: Yes, I thought that we wanted the user to insert a tolerance when working with Floating inputs. Do you want to set the default tolerance to something like `1e-14`?

ranocha: I think it would be more appropriate to use something based on the precision of the floating point numbers involved as mentioned above in https://github.com/ranocha/BSeries.jl/pull/156#discussion_r1315568089

Sondar74:
I will close this PR to start a clean one with only one tolerance as Keyword.

PR #161, opened 2023-09-05
Tolerances New PR

Sondar74:
New PR with tolerances.
Added absolute tolerance for `isapprox()` and exact comparison for `Rational` inputs.
New PR with one test for a RK method with Floating Inputs obtained from MatLab. Also, the non-broken tests for the AVF method passed.

### DIFF
--- a/src/BSeries.jl
+++ b/src/BSeries.jl
@@ -1748,16 +1748,18 @@ function renormalize!(series)
 end
 
 """
-    energy_preserving_order(rk::RungeKuttaMethod, max_order)
+    energy_preserving_order(rk::RungeKuttaMethod, max_order; tol::Real=0)
 
 This function checks up to which order a Runge-Kutta method `rk` is
 energy-preserving for Hamiltonian problems.
 It requires a `max_order` so that it does not run forever if the order up to
 which the method is energy-preserving is too big or infinite.
+Keyword Arguments:
+- `tol::Real=1e-14`: The absolute tolerance for energy preservation.
 
 See also [`is_energy_preserving`](@ref)
 """
-function energy_preserving_order(rk::RungeKuttaMethod, max_order)
+function energy_preserving_order(rk::RungeKuttaMethod, max_order; tol::Real=0)
     p = 0
     not_energy_preserving = false
     while not_energy_preserving == false
@@ -1766,7 +1768,7 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
             return max_order
         end
         # Check energy preservation up to the given order
-        if is_energy_preserving(rk, p + 1) == false
+        if is_energy_preserving(rk, p + 1, tol=tol) == false
             not_energy_preserving = true
         end
         p = p + 1
@@ -1775,22 +1777,26 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
 end
 
 """
-    is_energy_preserving(rk::RungeKuttaMethod, order)::Bool
+    is_energy_preserving(rk::RungeKuttaMethod, order; tol::Real=0)::Bool
 
 This function checks whether the Runge-Kutta method `rk` is
 energy-preserving for Hamiltonian systems up to a given `order`.
+Keyword Arguments:
+- `tol::Real=1e-14`: The absolute tolerance for energy preservation.
 """
-function is_energy_preserving(rk::RungeKuttaMethod, order)
+function is_energy_preserving(rk::RungeKuttaMethod, order; tol::Real=0)
     series = bseries(rk, order)
-    return is_energy_preserving(series)
+    return is_energy_preserving(series, tol=tol)
 end
 
 """
-    is_energy_preserving(series_integrator)::Bool
+    is_energy_preserving(series_integrator; tol::Real=0)::Bool
 
 This function checks whether the B-series `series_integrator` of a time
 integration method is energy-preserving for Hamiltonian systems - up to the
 [`order`](@ref) of `series_integrator`.
+Keyword Arguments:
+- `tol::Real=1e-14`: The absolute tolerance for energy preservation.
 
 # References
 
@@ -1800,7 +1806,7 @@ This code is based on the Theorem 2 of
   Foundations of Computational Mathematics 10 (2010): 673-693.
   [DOI: 10.1007/s10208-010-9073-1](https://link.springer.com/article/10.1007/s10208-010-9073-1)
 """
-function is_energy_preserving(series_integrator)
+function is_energy_preserving(series_integrator; tol::Real=0)
     # This method requires the B-series of a map as input
     let t = first(keys(series_integrator))
         @assert isempty(t)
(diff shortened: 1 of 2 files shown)


src/BSeries.jl
```diff
@@ -1991,8 +2008,19 @@ function _is_energy_preserving_sparse(trees, coefficients)
...
+                    return false
+                end
+            elseif tol != 0
+                if !isapprox(coefficients[t_index], 0, atol = tol)
+                    return false
+                end
+            else
+                if !iszero(coefficients[t_index])
+                    return false
+                end
```
ranocha, line 2023: I do not like that this code is duplicated. You could move it to a small function and call it from both `_is_energy_preserving_sparse` and `_is_energy_preserving_dense`.
Moreover, I think the way how you handle the default tolerance can be improved. 
- First, if the user selects `tol = 0`, we should respect that and use `tol = 0` - as described in the docstring.
- Next, there are much more floating point types - you could check `<: AbstractFloat` instead of `== Float64` etc.
- I gave you a concrete suggestion how to handle this in the previous PR:
  ```julia
  V = eltype(series_integrator)
  if tol === nothing
    if V <: AbstractFloat
      tol = 100 * eps(T)
    else
      tol = zero(T)
    end
  end
  ```
  _Originally posted in https://github.com/ranocha/BSeries.jl/pull/156#pullrequestreview-1609538103_
  Is there something wrong with such an approach in your opinion? To adapt it to this PR, you could create a new function, say `energy_preserving_default_tolerance(V, tol)`, that contains the logic above given the `valtype(series_integrator)` etc. This has to be documented appropriately in the docstrings of the high-level functions. Then, these high-level functions would have the default value `tol = nothing` that is converted appropriately by the helper function `energy_preserving_default_tolerance`.

test/runtests.jl
```diff
@@ -2115,6 +2115,18 @@ using Aqua: Aqua
             series = bseries(AverageVectorFieldMethod(BigFloat), 7)
             # TODO: This test is currently broken and throws an error
             @test_broken is_energy_preserving(series)
+
+            # Generated from MatLab
```
ranocha, line 2119: This does not really describe anything that helps people reading this code. Could you please describe how you generated it and what kind of properties it has?

ranocha (changes requested):
Thanks for your work!

> Also, the non-broken tests for the AVF method passed.

Do you mean that a previously broken test passes now? This is great! Could you please adapt the test accordingly?

src/BSeries.jl
Sondar74: ``````Okey, I can implement this function. Do you want to use the function `isapprox()` always then? I discussed with David about using `iszero()` for non - `AbstractFloat` inputs, but if we set a tolerance using this function
`function energy_preserving_default_tolerance(V, tol)
    if tol == 0
        if V <: AbstractFloat
            tol = 100 * eps(T)
        else
            tol = zero(T)
        end
    end
    return tol
end`
then we can use only the function `isapprox()` and only once.

ranocha: Yes, that's fine with me. It should simplify the control flow.

Sondar74: Okey. Please take a look at the new changes I just pushed and tell me what you think.

src/BSeries.jl
```diff
@@ -1748,16 +1748,19 @@ function renormalize!(series)
 end
 
 """
-    energy_preserving_order(rk::RungeKuttaMethod, max_order)
+    energy_preserving_order(rk::RungeKuttaMethod, max_order; tol::Real=0)
```
ranocha, line 1751: As I said, I would prefer using a default value that cannot be used as sensible value here - `nothing` instead of `0` to allow people to choose a vanishing tolerance.

test/runtests.jl
```diff
@@ -2115,6 +2115,21 @@ using Aqua: Aqua
...
+            A = [0.0000000000000000E+00	0.0000000000000000E+00	0.0000000000000000E+00	0.0000000000000000E+00	
+            -2.0314781009861151E-02	0.0000000000000000E+00	0.0000000000000000E+00	0.0000000000000000E+00	
+            1.0721250651661693E+01	-1.0078547105299583E+01	0.0000000000000000E+00	0.0000000000000000E+00	
+            -4.1299845612430317E+01	4.0239538957660457E+01	2.0603066547698594E+00	0.0000000000000000E+00]
+            b = [2.0623108969503225E+00	
+            -1.7306596124709785E+00	
+            5.6957376564633877E-01	
+            9.8774949874317328E-02	]
+            rk = RungeKuttaMethod(A,b)
+            @test energy_preserving_order(rk, 10) == 4
```
ranocha, line 2132: Could you please add a test for the `accuracy_order` of the method, too?

src/BSeries.jl
```diff
@@ -1775,22 +1778,28 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
...
 """
-    is_energy_preserving(series_integrator)::Bool
+    is_energy_preserving(series_integrator; tol::Real=0)::Bool
 
 This function checks whether the B-series `series_integrator` of a time
 integration method is energy-preserving for Hamiltonian systems - up to the
 [`order`](@ref) of `series_integrator`.
+Keyword Arguments:
+- `tol = nothing`: The absolute tolerance for energy preservation. If the elements of
+`rk` are Floating Points, the default tolerance is set to tol = 1e-14.
```
ranocha, line 1802: Could you please describe the actual implementation we use here?

src/BSeries.jl
```diff
@@ -1775,22 +1778,28 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
...
-function is_energy_preserving(rk::RungeKuttaMethod, order)
+function is_energy_preserving(rk::RungeKuttaMethod, order; tol=nothing)
     series = bseries(rk, order)
-    return is_energy_preserving(series)
+    return is_energy_preserving(series, tol=tol)
 end
 
 """
-    is_energy_preserving(series_integrator)::Bool
+    is_energy_preserving(series_integrator; tol::Real=0)::Bool
```
ranocha, line 1795: This should be updated

src/BSeries.jl
```diff
@@ -1775,22 +1778,28 @@ function energy_preserving_order(rk::RungeKuttaMethod, max_order)
...
 """
-    is_energy_preserving(rk::RungeKuttaMethod, order)::Bool
+    is_energy_preserving(rk::RungeKuttaMethod, order; tol::Real=0)::Bool
 
 This function checks whether the Runge-Kutta method `rk` is
 energy-preserving for Hamiltonian systems up to a given `order`.
+Keyword Arguments:
+- `tol = nothing`: The absolute tolerance for energy preservation. If the elements of
+`rk` are Floating Points, the default tolerance is set to tol = 1e-14.
 """
```
ranocha, line 1788: See below

src/BSeries.jl
```diff
@@ -1748,16 +1748,19 @@ function renormalize!(series)
...
-    energy_preserving_order(rk::RungeKuttaMethod, max_order)
+    energy_preserving_order(rk::RungeKuttaMethod, max_order; tol::Real=0)
 
 This function checks up to which order a Runge-Kutta method `rk` is
 energy-preserving for Hamiltonian problems.
 It requires a `max_order` so that it does not run forever if the order up to
 which the method is energy-preserving is too big or infinite.
+Keyword Arguments:
+- `tol = nothing`: The absolute tolerance for energy preservation. If the elements of
+`rk` are Floating Points, the default tolerance is set to tol = 1e-14.
```
ranocha, line 1759: See below

ranocha (changes requested):
Thanks a lot for your patience. We're almost there

PR #210, opened 2024-06-06
fix conflicting exports

ranocha:


ranocha:
The documentation error is caused by https://github.com/SciML/OrdinaryDiffEq.jl/issues/2182 and needs a new relase of RecursiveArrayTools.jl

PR #257, opened 2025-02-01
Update bseries_creation.md

goodship1:


ketch:
This is technically correct but frankly I don't like cluttering the git history with insignificant one- or two-character modifications.  If we are worried about things like this then we could just run a grammar- and spell-checker over the whole repo and do it all in one commit.

PR #299, opened 2025-11-12
Add support for Two-Derivative RK Methods

JohnDriscollAcademic:
Adds support for two-derivative and additive two-derivative Runge-Kutta methods, including the necessary structures and functions to work with B-series. Note I defined the structures directly where the existing methods seem to import them from RootedTrees.jl.

### DIFF
--- a/src/BSeries.jl
+++ b/src/BSeries.jl
@@ -1278,13 +1278,6 @@ struct TwoDerivativeRungeKuttaMethod{T,
     b2::VecT
 end
 
-"""
-    TwoDerivativeRungeKuttaMethod(A1, b1, A2, b2, c1 = vec(sum(A1, dims=2)))
-
-Construct a two-derivative Runge–Kutta method from coefficient arrays.
-All inputs are promoted to a common numeric type.  
-`c1` defaults to the usual row-sum condition `sum(A1, dims=2)`.
-"""
 function TwoDerivativeRungeKuttaMethod(A1, b1, A2, b2, c1 = vec(sum(A1, dims=2)))
     # promote all numeric types together
     T = promote_type(eltype(A1), eltype(b1), eltype(A2), eltype(b2), eltype(c1))
--- a/src/BSeries.jl
+++ b/src/BSeries.jl
@@ -1093,6 +1093,333 @@ function RootedTrees.derivative_weight(t::RootedTree,
     return result
 end
 
+
+#####################################################################
+# defining the new methods, not sure where best place for this code is.
+
+"""
+    TwoDerivativeRungeKuttaMethod{T, RKm} <: AbstractTimeIntegrationMethod
+
+Two-derivative Runge–Kutta (TDRK) method composed of multiple
+`RungeKuttaMethod`s defining base and derivative components.
+"""
+struct TwoDerivativeRungeKuttaMethod{T, RKm <: AbstractVector{<:RungeKuttaMethod{T}}} <: RootedTrees.AbstractTimeIntegrationMethod
+    rkm::RKm
+end
+
+"""
+    TwoDerivativeRungeKuttaMethod(rkm)
+
+Construct from a vector of `RungeKuttaMethod`s.  
+Promotes all coefficient types to a common numeric type.
+"""
+function TwoDerivativeRungeKuttaMethod(rkm)
+    T = mapreduce(eltype, promote_type, rkm)           # promote coefficient type
+    As = map(rk -> T.(rk.A), rkm)
+    bs = map(rk -> T.(rk.b), rkm)
+    cs = map(rk -> T.(rk.c), rkm)
+    return TwoDerivativeRungeKuttaMethod(As, bs, cs)
+end
+
+"""
+    TwoDerivativeRungeKuttaMethod(As, bs[, cs])
+
+Construct directly from coefficient arrays.  
+If c not explicitly given, computed as `sum(A, dims = 2)` per usual.
+"""
+function TwoDerivativeRungeKuttaMethod(As, bs, cs = map(A -> vec(sum(A, dims = 2)), As))
+    rkm = map(RungeKuttaMethod, As, bs, cs)
+    return TwoDerivativeRungeKuttaMethod(rkm)
+end
+
+"""
+    eltype(tdrk::TwoDerivativeRungeKuttaMethod)
+
+Return the element type of the coefficients.
+"""
+Base.eltype(tdrk::TwoDerivativeRungeKuttaMethod{T}) where {T} = T
+
+
+"""
+    TwoDerivativeAdditiveRungeKuttaMethod{T, RKm} <: AbstractTimeIntegrationMethod
+
+Additive two-derivative Runge–Kutta (TDARK) method composed of multiple
+`RungeKuttaMethod`s, typically representing `(f, f′, g)`.
+"""
+struct TwoDerivativeAdditiveRungeKuttaMethod{T, RKm <: AbstractVector{<:RungeKuttaMethod{T}}} <: RootedTrees.AbstractTimeIntegrationMethod
+    rkm::RKm
+end
+
+"""
+    TwoDerivativeAdditiveRungeKuttaMethod(rkm)
+
+Construct from a vector of `RungeKuttaMethod`s (e.g., `(f, f′, g)`).
+Promotes coefficient types to a common numeric type.
+"""
+function TwoDerivativeAdditiveRungeKuttaMethod(rkm)
+    T = mapreduce(eltype, promote_type, rkm)     # promote element type
+    As = map(rk -> T.(rk.A), rkm)
+    bs = map(rk -> T.(rk.b), rkm)
+    cs = map(rk -> T.(rk.c), rkm)
+    return TwoDerivativeAdditiveRungeKuttaMethod(As, bs, cs)
+end
+
+"""
+    TwoDerivativeAdditiveRungeKuttaMethod(As, bs[, cs])
+
+Construct directly from coefficient arrays.  
+"""
+function TwoDerivativeAdditiveRungeKuttaMethod(As, bs, cs = map(A -> vec(sum(A, dims = 2)), As))
+    rkm = map(RungeKuttaMethod, As, bs, cs)
+    return TwoDerivativeAdditiveRungeKuttaMethod(rkm)
+end
+
+"""
+    eltype(tdark::TwoDerivativeAdditiveRungeKuttaMethod)
+
+Return the coefficient element type.
+"""
+Base.eltype(tdark::TwoDerivativeAdditiveRungeKuttaMethod{T}) where {T} = T
+
+#####################################################################
+
+"""
+    elementary_weight(t::ColoredRootedTree, tdrk::TwoDerivativeRungeKuttaMethod) -> Number
+
+Compute the elementary weight associated with the colored rooted tree `t`
+for a two-derivative Runge–Kutta method `tdrk`.
+
+This function first generates all possible collapsed forms of `t` using
+[`collapse_tree`](@ref), then sums their contributions weighted by their
+multiplicities. Each tree’s contribution is obtained from [`tree_weight`](@ref).
+
+
+This follows from the Butcher type order conditions exhibited in, 
+Chan, R.P.K., Tsai, A.Y.J. On explicit two-derivative Runge-Kutta methods.
+- Numer Algor 53, 171–194 (2010). https://doi.org/10.1007/s11075-009-9349-1
+ 
+
+# Arguments
+- `t`: A `ColoredRootedTree` representing the current term.
+- `tdrk`: The `TwoDerivativeRungeKuttaMethod` whose coefficients define the
+  weights.
+
+# Returns
+A scalar weight equal to the sum over all collapsed trees.
+"""
+function elementary_weight(t::ColoredRootedTree, tdrk::TwoDerivativeRungeKuttaMethod)
+    trees, multiplicities = collapse_tree(t)
+
+    sum = zero(eltype(multiplicities))  # accumulator for total weight
+    for (i, tree) in enumerate(trees)  # sum over tree and all collapsed forms
+        sum += multiplicities[i] * tree_weight(tree, tdrk)
+    end
+    return sum
+end
+
+
+"""
+    tree_weight(t::ColoredRootedTree, tdrk::TwoDerivativeRungeKuttaMethod) -> Number
+
+Compute the weight contribution of a single colored rooted tree `t`
+for the two-derivative Runge–Kutta method `tdrk`.
+
+The function recursively evaluates the contribution of each subtree
+using the appropriate `(A, b, c)` coefficients based on node color:
+
+- Color `1` → derivative part, F.           (uses `a1`, `b1`, `c1`)
+- Color `2` → second derivative part, F'.   (uses `a2`, `b2`, `c2`)
+
+# Arguments
+- `t`: A `ColoredRootedTree`.
+- `tdrk`: A `TwoDerivativeRungeKuttaMethod` containing two embedded RK tableau.
+
+# Returns
+A scalar weight equal to the B-series coefficient associated with the tree.
+"""
+function tree_weight(t::ColoredRootedTree, tdrk::TwoDerivativeRungeKuttaMethod)
+    b1 = (tdrk.rkm)[1].b  # regular b coefficients
+    b2 = (tdrk.rkm)[2].b  # derivative b coefficients
+    a1 = (tdrk.rkm)[1].A  # regular A coefficients
+    a2 = (tdrk.rkm)[2].A  # derivative A coefficients
+    c1 = (tdrk.rkm)[1].c  # regular c coefficients
+    c2 = (tdrk.rkm)[2].c  # derivative c coefficients
+
+    level_sequence = t.level_sequence
+    color_sequence = t.color_sequence
+
+    # Choose root b vector based on root color
+    b_root = color_sequence[1] == 1 ? b1 : b2
+
+    # Recursive evaluation of the product of subtree contributions
+    function helper(t::ColoredRootedTree)
+        # Compute contribution of a subtree
+        function subtree_contribution(t::ColoredRootedTree)
+            A, c = t.color_sequence[1] == 1 ? (a1, c1) : (a2, c2)
+
+            # Leaf node contributes only its c-value
+            if length(t.level_sequence) == 1
+                return c
+            else
+                # Internal node: multiply A by recursive subtree product
+                return A * helper(t)
+            end
+        end
+
+        # Initialize product vector (same length as b1)
+        product = ones(eltype(b1), length(b1))
+
+        # Get subtrees of the current node
+        subtrees_array = subtrees(t)
+
+        # If no subtrees, return ones vector
+        if isempty(subtrees_array)
+            return product
+        else
+            # Multiply elementwise by each subtree’s contribution
+            for n in subtrees_array
+                product = product .* subtree_contribution(n)
+            end
+        end
+
+        return product
+    end
+
+    # Final weight = b_root ⋅ product of subtree contributions
+    product = helper(t)
+    result = LinearAlgebra.dot(b_root, product)
+    return result
+end
+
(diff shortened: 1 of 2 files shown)


ranocha (changes requested):

src/BSeries.jl
```diff
@@ -1093,6 +1093,332 @@ function RootedTrees.derivative_weight(t::RootedTree,
     return result
 end
 
+#####################################################################
+# defining the new methods, not sure where best place for this code is.
```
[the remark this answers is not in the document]
JohnDriscollAcademic: Thanks for the suggestions. I have cut back to just focusing on the standard two-derivative methods for this pr, put all the new code in one section, and added a test. I am not very experienced with contributing on GitHub, so I have changed this to a "Draft" Pull Request in case more iteration is needed.

Regarding the test, so far I have just had it use a known method from the paper by Robert Chan, and had it check that the proper `TwoDerivativeRungeKuttaMethod` struct was created, that the B-series computed the correct order of accuracy for the method, and checked a specific tree weight. This seems to be in line with what other tests have done, but please let me know if anything else should be included.

src/BSeries.jl
```diff
@@ -1253,6 +1255,334 @@ end
...
+    series[rootedtree(Int[])] = one(V)
+
+    # iterate over orders
+    for o in 1:order
+        for t in RootedTreeIterator(o)
+            color_sequence = fill(1, length(t.level_sequence))
+            gen_t = rootedtree(t.level_sequence, color_sequence)
+            # assign coefficient via elementary weight (only def for colored trees)
+            series[copy(t)] = collapse_elementary_weight(gen_t, tdrk)
+        end
```
ranocha: I am not sure whether I like this approach. A two-derivative RK method should have a B-series expansion in terms of standard rooted trees, not colored trees.

ranocha (commented):
Thanks! It looks like your approach will already allocate some temporary arrays. Could you please compare the performance and results with an approach creating the B-series using `compose` as in https://ranocha.de/BSeries.jl/stable/tutorials/bseries_creation/#tutorial-bseries-creation-RK-compose?

This should work directly for an explicit method. For an implicit method, it should work when used as a fixed point map with enough iterations so that the result does not change anymore up to the desired `order`. This approach is used for Runge-Kutta methods in Section 313 of Butcher's book "Numerical Methods for Ordinary Differential Equations" to derive the order conditions of RK methods.

test/runtests.jl
```diff
@@ -3245,4 +3245,57 @@ using Aqua: Aqua
             @test substituted == s
         end
     end
+
+    @testset "TwoDerivativeRungeKuttaMethod" begin
+
+        # Example two-derivative RK method (order 3) from "On explicit two-derivative Runge-Kutta methods" from R.P.K. Chan and A.Y.J. Tsai (2010)
+        A1 = [0 0;
```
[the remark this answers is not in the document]
JohnDriscollAcademic: I knew my implementation was far from memory optimized but I did not think it was this bad. I did some quick benchmarking and here are the results. For my method, collapse all trees, then find order conditions from there, I get the following where I just test the time it takes to generate the B-series for a given order:

        Computing the B-series coefficients up to order: 7
        0.009580 seconds (291.81 k allocations: 14.426 MiB)
        0.018783 seconds (291.64 k allocations: 14.417 MiB, 49.28% gc time)
        
        Computing the B-series coefficients up to order: 8
        0.061796 seconds (1.44 M allocations: 72.963 MiB, 15.03% gc time)
        0.054118 seconds (1.44 M allocations: 72.963 MiB, 13.11% gc time)
        
        Computing the B-series coefficients up to order: 9
        0.295879 seconds (7.27 M allocations: 379.771 MiB, 18.92% gc time)
        0.290079 seconds (7.27 M allocations: 379.771 MiB, 16.50% gc time)

where I’ve run twice to account for Julia's compilation time.

Calculating differentials the more streamlined way, that is as recursively defined in the paper and implemented in the thesis you shared with me, we get far nicer results, polynomial rather than exponential!!

      Computing the B-series coefficients up to order: 7
      0.005005 seconds (76.89 k allocations: 3.987 MiB)
      0.004866 seconds (76.86 k allocations: 3.985 MiB)
      
      Computing the B-series coefficients up to order: 8
      0.020274 seconds (272.40 k allocations: 14.138 MiB)
      0.019488 seconds (272.40 k allocations: 14.138 MiB)
      
      Computing the B-series coefficients up to order: 9
      0.087649 seconds (968.06 k allocations: 50.192 MiB, 28.02% gc time)
      0.063105 seconds (968.06 k allocations: 50.192 MiB)

This shown, I think the recursive method is obviously what I should have implemented. Why I originally veered away from it was my goal was generating order conditions for the low-stage IMEX multi-derivative method (which now I know is possible only since we went to order 4 :). It's not immediately clear to me how to generate the weights recursively for the additive case, however with some thinking I’m sure we could figure it out. Additionally, doing things by generating trees also allowed me to quickly reduce the order conditions in the case that either component was linear; it's also not clear to me how easy this would be to implement in the recursive case.

All that said, if you would like to move forward with multi-derivative methods, I could push a version using the recursive definition, which I just tested as agreeing with the 5s7p method. That said, my implementation is almost line for line with that of the thesis, so if we were to merge something like that I would hope that student would be properly credited. As for multi-derivative additive methods, I would need to do some more thinking before reimplementing that in the memory-friendly manner.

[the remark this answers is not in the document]
JohnDriscollAcademic: Apologies for the late reply, things got busy with our Thanksgiving holiday followed by the start of our exam period. I have updated the PR to include the more efficient method and added a test for a method up to order 5 from the referenced paper. I'll try benchmarking the current implementation again; while it certainly outperformed my original implementation, it will be nice to see how it compares to something like the standard or additive methods existing.

src/BSeries.jl
```diff
@@ -1253,6 +1255,334 @@ end
...
+
+"""
+    TwoDerivativeRungeKuttaMethod(A1, b1, A2, b2, c = vec(sum(A1, dims=2)))
+
+Represent a two-derivative Runge-Kutta method with Butcher coefficients
+`A1`, `b1`, and `c` for the first derivative and `A2`, `b2` for the second
+derivative.
+If `c` is not provided, the usual "row sum" requirement of consistency with
+autonomous problems is applied.
+```
```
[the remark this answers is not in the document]
JohnDriscollAcademic: I added this type of TeX expression, but Julia didn’t like the \ since it’s a string escape char, so I just had to change all \ --> \\,  I’m not sure how this would affect the rendering if it’s automatically compiled by documentation or something like that.

test/runtests.jl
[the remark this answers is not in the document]
ranocha: That would be nice indeed 👍

src/BSeries.jl
```diff
@@ -1254,33 +1256,231 @@ end
...
-- Philippe Chartier, Ernst Hairer, Gilles Vilmart (2010)
-  Algebraic Structures of B-series.
-  Foundations of Computational Mathematics
-  [DOI: 10.1007/s10208-010-9065-1](https://doi.org/10.1007/s10208-010-9065-1)
 """
-function substitute(b, a, t::AbstractRootedTree)
-    result = zero(first(values(a)) * first(values(b)))
+struct TwoDerivativeRungeKuttaMethod{T,
+                                     MatT <: AbstractMatrix{T},
+                                     VecT <: AbstractVector{T}} <:RootedTrees.AbstractTimeIntegrationMethod
```
ranocha, line 1287: It looks like the method `substitute(b, a, t::AbstractRootedTree)` has been deleted here?

src/BSeries.jl
```diff
@@ -1254,33 +1256,231 @@ end
...
+`A1`, `b1`, and `c` for the first derivative and `A2`, `b2` for the second
+derivative.
+If `c` is not provided, the usual "row sum" requirement of consistency with
+autonomous problems is applied.
 
-# References
+Given an ODE ``u'(t) = f(t, u(t))`` with ``u''(t) = g(t, u(t))``,
+one step from ``u^{n}`` to ``u^{n+1}`` is given by
+
+'''math
```
ranocha:
````suggestion
```math
````

File: src/BSeries.jl [hunk 7]
```diff
@@ -1254,33 +1256,231 @@ end
...
+\\begin{aligned}
+  y^i &= u^n + \\Delta t \\sum_j a^{1}_{i,j} f(t^n + c_i \\Delta t, y^i) + \\Delta t^2 \\sum_j a^{2}_{i,j} g(t^n + c_i \\Delta t, y^i), \\\
+  u^{n+1} &= u^n + \\Delta t \\sum_i b^1_{i} f(t^n + c_i \\Delta t, y^i) + \\Delta t^2 \\sum_i b^2_{i} g(t^n + c_i \\Delta t, y^i),
+\\end{aligned}
+'''
```
ranocha:
````suggestion
```
````

File: src/BSeries.jl [hunk 8]
```diff
@@ -1254,33 +1256,231 @@ end
...
+    TwoDerivativeRungeKuttaMethod(A1, b1, A2, b2, c = vec(sum(A1, dims=2)))
 
-Compute the coefficient corresponding to the tree `t` of the B-series that is
-formed by substituting the B-series `b` into the B-series `a`. It is assumed
-that the B-series `b` has the coefficient zero of the empty tree.
+Represent a two-derivative Runge-Kutta method with Butcher coefficients
+`A1`, `b1`, and `c` for the first derivative and `A2`, `b2` for the second
+derivative.
+If `c` is not provided, the usual "row sum" requirement of consistency with
+autonomous problems is applied.
```
ranocha, line 1265: Can you please add some references (following the format we have used before) to the end of the docstring?

src/BSeries.jl
```diff
@@ -1254,33 +1256,231 @@ end
...
-    result = zero(first(values(a)) * first(values(b)))
+struct TwoDerivativeRungeKuttaMethod{T,
+                                     MatT <: AbstractMatrix{T},
+                                     VecT <: AbstractVector{T}} <:RootedTrees.AbstractTimeIntegrationMethod
+    A1::MatT
+    b1::VecT
+    c1::VecT
+    A2::MatT
+    b2::VecT
+    c2::VecT
```
ranocha: The documentation says we have only

```suggestion
    b1::VecT
    c::VecT
    A2::MatT
    b2::VecT
```

test/runtests.jl
```diff
@@ -3245,4 +3245,71 @@ using Aqua: Aqua
...
+        b1 = [1, 0]
+
+        A2 = [0 0;
+            1//8 0]
+
+        b2 = [1//6, 1//3]
+
+        #constructor
+
+        tdrk = TwoDerivativeRungeKuttaMethod(A1, b1, A2, b2)
```
ranocha:
```suggestion
        tdrk = @inferred TwoDerivativeRungeKuttaMethod(A1, b1, A2, b2)
```

To check type stability

test/runtests.jl
```diff
@@ -3245,4 +3245,71 @@ using Aqua: Aqua
...
+
+        b_2 = [
+            71//960,
+            2401//4800,
+            -625//1728,
+            2401//8640,
+            13//1350
+        ]
+
+        tdrk_2 = TwoDerivativeRungeKuttaMethod(A_1, b_1, A_2, b_2)
```
ranocha:
```suggestion
        tdrk_2 = @inferred TwoDerivativeRungeKuttaMethod(A_1, b_1, A_2, b_2)
```

ranocha (changes requested):

src/BSeries.jl
JohnDriscollAcademic: Sorry, don't know how that happened, fixing now!

src/BSeries.jl
JohnDriscollAcademic: In general c2 should always be zero, I'm not sure if its best to hard code that in, or still leave the option for it to be the row sum of A2

test/runtests.jl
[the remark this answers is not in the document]
JohnDriscollAcademic: This new code seems reasonable performance-wise. In the following I generate the B-series of various order for various five-stage methods.

For Two-Derivative,

      Computing the B-series coefficients up to order: 7
      0.030486 seconds (76.89 k allocations: 4.090 MiB, 82.09% GC time)
      0.004316 seconds (76.86 k allocations: 4.088 MiB)
      
      Computing the B-series coefficients up to order: 8
      0.016571 seconds (272.40 k allocations: 14.510 MiB)
      0.016692 seconds (272.40 k allocations: 14.510 MiB)
      
      Computing the B-series coefficients up to order: 9
      0.059871 seconds (968.06 k allocations: 51.528 MiB)
      0.078173 seconds (968.06 k allocations: 51.528 MiB, 23.73% GC time)

For Standard RK, using the standard (not derivative part) Butcher table from the two-derivative example,

      Computing the B-series coefficients up to order: 7
      0.000456 seconds (4.07 k allocations: 289.219 KiB)
      0.000436 seconds (4.07 k allocations: 289.219 KiB)
      
      Computing the B-series coefficients up to order: 8
      0.001130 seconds (11.21 k allocations: 809.773 KiB)
      0.001265 seconds (11.21 k allocations: 809.773 KiB)
      
      Computing the B-series coefficients up to order: 9
      0.003318 seconds (31.23 k allocations: 2.161 MiB)
      0.003326 seconds (31.23 k allocations: 2.161 MiB)

And existing AdditiveRungeKuttaMethod with a 3p4s IMEX method from https://doi.org/10.1016/S0168-9274(97)00056-1 ,


      Computing the B-series coefficients up to order: 7
      0.005859 seconds (285.18 k allocations: 13.408 MiB)
      0.014277 seconds (285.18 k allocations: 13.408 MiB, 57.02% GC time)
      
      Computing the B-series coefficients up to order: 8
      0.048015 seconds (1.50 M allocations: 70.561 MiB, 30.28% GC time)
      0.048981 seconds (1.50 M allocations: 70.561 MiB, 33.61% GC time)
      
      Computing the B-series coefficients up to order: 9
      0.505602 seconds (7.97 M allocations: 379.642 MiB, 63.28% GC time)
      0.532601 seconds (7.97 M allocations: 379.642 MiB, 63.10% GC time)

That is to say, two-derivative is outperforming additive and underperforming standard RK, which seems reasonable.

src/BSeries.jl
ranocha: If it has to be zero, I would like to avoid including them to make the code easier to follow and reduce the number of operations. Adding comments explaining this would be nice.
However, if the code can be simplified by including `c2`, please do so. Just keep the documentation and the code consistent, please.

test/runtests.jl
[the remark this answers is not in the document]
ranocha: 👍

src/BSeries.jl
JohnDriscollAcademic: Yes, the c's are confusing to me. In the paper's recursive definition of the order conditions, only c is written explicitly, and it is stated to be the standard row sum A1 * e of the primary tableau. However, in this setting there is also a c_hat (here c2), given by A2 * e from the derivative tableau. This quantity does appear computationally in the order conditions through the multiplication A2 * e itself.

I would assume that one could formulate everything, in the sense of calculating tree weights, without introducing any c at all, always interpreting it implicitly as A * e for the appropriate tableau. My understanding is that the paper writes c explicitly because some readers may wish to deviate from the standard choice c = A1 * e, whereas c2 is always taken to be A2 * e. That said, I will revert to using just one c as the row sum of the main tableau, since that runs the same as before and is slightly simpler._

src/BSeries.jl
```diff
@@ -1253,6 +1255,244 @@ end
...
+derivative.
+If `c` is not provided, the usual "row sum" requirement of consistency with
+autonomous problems is applied.
+
+Given an ODE ``u'(t) = f(t, u(t))`` with ``u''(t) = g(t, u(t))``,
+one step from ``u^{n}`` to ``u^{n+1}`` is given by
+
+```math
+\\begin{aligned}
+  y^i &= u^n + \\Delta t \\sum_j a^{1}_{i,j} f(t^n + c_i \\Delta t, y^i) + \\Delta t^2 \\sum_j a^{2}_{i,j} g(t^n + c_i \\Delta t, y^i), \\\
```
ranocha:
```suggestion
  y^i &= u^n + \\Delta t \\sum_j a^{1}_{i,j} f(t^n + c_j \\Delta t, y^j) + \\Delta t^2 \\sum_j a^{2}_{i,j} g(t^n + c_j \\Delta t, y^j), \\\
```

src/BSeries.jl
```diff
@@ -1253,6 +1255,244 @@ end
...
+  u^{n+1} &= u^n + \\Delta t \\sum_i b^1_{i} f(t^n + c_i \\Delta t, y^i) + \\Delta t^2 \\sum_i b^2_{i} g(t^n + c_i \\Delta t, y^i),
```
ranocha:
```suggestion
  u^{n+1} &= u^n + \\Delta t \\sum_i b^1_{i} f(t^n + c_i \\Delta t, y^i) + \\Delta t^2 \\sum_i b^2_{i} g(t^n + c_i \\Delta t, y^i).
```

src/BSeries.jl
```diff
@@ -1253,6 +1255,244 @@ end
...
+\\end{aligned}
+```
+
+# References
+
+-Chan, R.P.K., Tsai, A.Y.J. 
+"On explicit two-derivative Runge-Kutta methods."
+Numer Algor 53, 171–194 (2010):
+[DOI: 10.1007/s11075-009-9349-1](https://doi.org/10.1007/s11075-009-9349-1)
```
ranocha:
```suggestion
- Chan, R.P.K., Tsai, A.Y.J. 
  "On explicit two-derivative Runge-Kutta methods."
  Numer Algor 53, 171–194 (2010):
  [DOI: 10.1007/s11075-009-9349-1](https://doi.org/10.1007/s11075-009-9349-1)
```

src/BSeries.jl
ranocha: Did you build the docs locally? Three backslashes look suspicious to me.

src/BSeries.jl
```diff
@@ -1253,6 +1255,244 @@ end
...
+- Numer. Algor 53, 171–194 (2010). https://doi.org/10.1007/s11075-009-9349-1
+
+#see formula 16 in the paper
+# alpha(t) = b1*eta(subtrees(t)) +b2*eta(collapse_trees(nu))
+
+# Arguments
+- `t`: A `RootedTree` representing the current term.
+- `tdrk`: The `TwoDerivativeRungeKuttaMethod` whose coefficients define the
+  weights.
+
```
ranocha: Please clean up code/docstring formatting

src/BSeries.jl
JohnDriscollAcademic: Sorry, I hadn’t rebuilt the docs after adding the extra backslash, which I do think should be removed. I’m also having some trouble understanding how LaTeX works in Julia docstrings.

One odd issue I’ve noticed is in the docstring:

    Given an ODE ``u'(t) = f(t, u(t))`` with ``u''(t) = g(t, u(t))``,
    one step from ``u^{n}`` to ``u^{n+1}`` is given by

The double quotes after u for the second derivative behave fine in a code editor, yet on GitHub they appear as escape characters, which seems to terminate the docstring and makes the rest of the code hard to read.

More importantly, when I just tried to build the docs, I got the following error from the doctest of continuous-stage methods:

    │ ERROR: MethodError: no method matching elementary_weight(::RootedTree{Int64, Vector{Int64}}, ::ContinuousStageRungeKuttaMethod{Matrix{Rational{Int64}}})
    │ The function `elementary_weight` exists, but no method is defined for this combination of argument types.
    │ Closest candidates are:
    │   elementary_weight(::RootedTree, !Matched::TwoDerivativeRungeKuttaMethod)
    │    @ BSeries ~/BSeries/src/BSeries.jl:1362

Does the fact that for continuous methods it is defining the function as `rootedtree.elementary_weight` versus for TDRK it is `elementary_weight` directly affect type dispatch? All the tests ran fine for continuous stage Runge-Kutta methods with this new code so I was surprised to see it here.

